PR #3, opened 2018-07-03
pull build from lastest sourceforge to git

ChristinaXu2017:
update build file to latest sourceforge


ChristinaXu2017:
debug file run.sh is deleted

PR #4, opened 2018-07-19
updated to include alt in equals method, added tests

holmeso:
Updated equals and hashcode so that alt is included, made immutable, and finally implements Comparable for ease of sorting.
Added some tests too!


holmeso:
I think its ok, but will plonk some code in to deal with that anyway

holmeso:
performs compareTo on alt should compareTo on ChrPositionName return 0

delocalizer:
Sorry, now that I look at it the `cpRef.compareTo` looks like it's inherited from https://github.com/AdamaJava/adamajava/blob/79b3d677c01e0b58ef35860470a1a2255c7c8261/qcommon/src/org/qcmg/common/model/ChrRangePosition.java#L105 which doesn't even compare about the reference allele.

delocalizer:
So if you're just comparing positions for `ChrPositionName` then certainly alt doesn't come into it for `ChrPositionRefAlt`

holmeso:
Have added ref and alt to compareTo method. Tests have been updated and extended

PR #13, opened 2018-08-09
Add parameters to confidence mode

holmeso:
Changes to the annotation process to more align the process to what the current production pipeline does.
Have paramaterised the values that have been changed so that we can move back to those values without (java) code changes should that ever be desirable (will need to update wdl to take advantage of this).


holmeso:
Details of results of these changes can be found here:
https://genomiqa.atlassian.net/wiki/spaces/PROJ/pages/31686657/HiSeq+Overlaps

PR #25, opened 2018-12-10
Values from vcf filter fields will now appear in maf

### LINKED ISSUE #9: Vcf2Maf filter field not being set
FT format field is now used to convey filter information for a sample, and this is not making it into the maf files.
Previous version of the code would have populated the high-level filter field in the vcf record and put this into the maf record.
Should all FT format fields be put into the maf or just the ones that were used to determine whether it is a PASS/FAIL? (ie. test if somatic, control otherwise)

holmeso:
In the olden days, the `QFlag` maf field would be populated from the vcf filter field.
This was great when our tools (`qsnp`, `qannotate`) populated the vcf filter field.

In this brave new world, there is a filter field for each sample column in a vcf record, and so, `qannotate`'s `Vcf2maf` mode has been updated to use the data from these sample filter fields (`FT`) to populate the `QFlag` field.

Test have also been added

fixes #9


holmeso:
It is expected that this will cause the nightly regression (latest) to fail as an `md5sum` of the all maf file is performed.

holmeso:
failing test - investigating...

holmeso:
tests are now passing

PR #29, opened 2018-12-19
update qcommon, qprofiler readay for new qprofiler2

ChristinaXu2017:
move some constants from qcommon to qprofier since they are old qprofiler only, not common any more


ChristinaXu2017:
stop merge since error

PR #30, opened 2018-12-19
Qprofiler2 develop

ChristinaXu2017:
tidy up code ready for qprofiler2;  move qprofiler1 only code from qcommon to qprofiler


ChristinaXu2017:
require merge again. the previous build failure caused by setting or accident click. want to try again

ChristinaXu2017:
update unit

holmeso:
Hi Christina, I'm part way through looking at your code.
Couple of points, would it be possible to try and adhere to [Googles java code styling](https://google.github.io/styleguide/javaguide.html)?
Eg. in `qcommon/src/org/qcmg/common/model/QCMGAtomicLongArray.java` you have a `if` nested inside a `for` with no braces

Also, looks like there is a bug in `qvisualise2/src/org/qcmg/qvisualise2/report/ReportBuilder.java`
Could you please add a test to check for this and fix.

ChristinaXu2017:
Qvisualise2 is paused at moment, it won’t work with new qprofiler2, I have to rewrite or big change it. I don’t think we should add any unit test for it.
 
I will try to follow the google styling, personally, I like short and fewer lines in codes. But I am happy to make changes if it makes the code more readable

PR #35, opened 2019-01-08
removed unnecessary null check of acc in getFormatFilter

### LINKED ISSUE #34: unnecessary null check
In qsnp's `GenotypeUtil` class, there is a redundant check for null in the `getFormatFilter` method.
The parameter `acc` is checked to see if it is null initially. It is then checked again a few lines later, which is unnecessary

The expected behaviour is that `acc` should only be checked for null-ness once

sarahk2019:
fixes #34 
acc argument is already checked that it is not null in the first line so changed initialisation of totalReadCount variable to not check for null acc, and removed null check of acc a few lines down.

### DIFF
--- a/qsnp/src/org/qcmg/snp/util/GenotypeUtil.java
+++ b/qsnp/src/org/qcmg/snp/util/GenotypeUtil.java
@@ -38,7 +38,7 @@ public class GenotypeUtil {
 		} else {
 			
 			boolean switchToSBiasCov = runSBias && ! AccumulatorUtils.bothStrandsByPercentage(acc, sbiasPercentageCoverage);
-			int totalReadCount = null != acc?  acc.getCoverage() : 0;
+			int totalReadCount = acc.getCoverage();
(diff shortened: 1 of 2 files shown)


delocalizer:
Eep I was too slow... Looks like the new version of `GenotypeUtil.java` is full of Windows line endings

holmeso:
where do you see the Windows line endings? It looks ok when viewing changes and looking at the file (normal and raw).
Build seems happy too...

delocalizer:
https://github.com/AdamaJava/adamajava/pull/35/files
And git diff

delocalizer:
475 lines removed, 478 lines added

delocalizer:
@holmeso @sarahk2019  actually the dos line endings _are_ in now master version; I just hadn't pull'ed the changes since branch was merged in. This is actually sort of good news, since it means there isn't some undocumented magic "line ending fixer-upper" being applied.

PR #45, opened 2019-01-24
Qprofiler2 develop

ChristinaXu2017:
merged qprofiler2 with new xml typle to this branch. I also convert all unit tests to junit4 rather than junit4; slightly fixed/upgrade java code. such as using "0" to replace null value since there is no counts.


ChristinaXu2017:
sorry for the typo. I mean current JUnit test is junit4, but not junit5.

Issue #48, opened 2019-02-02
SignatureGenerator dies with StackOverflow


delocalizer:
Preconditions:
Input bam contains just one `@SQ` reference contig in the header but SNP positions file has records at many different chromosomes. Untested, but probably also happens even when the input bam simply has fewer (not just one) `@SQ` contigs than the SNP positions file.

Result:
```
        java.lang.StackOverflowError
        ...
        at org.qcmg.sig.SignatureGeneratorBespoke.match(SignatureGeneratorBespoke.java:526)
        at org.qcmg.sig.SignatureGeneratorBespoke.match(SignatureGeneratorBespoke.java:527)
        at org.qcmg.sig.SignatureGeneratorBespoke.match(SignatureGeneratorBespoke.java:527)
        at org.qcmg.sig.SignatureGeneratorBespoke.match(SignatureGeneratorBespoke.java:527)
        at org.qcmg.sig.SignatureGeneratorBespoke.match(SignatureGeneratorBespoke.java:527)
```

So this was an interesting one... it comes down to the behaviour of `ChrPositionComparator.getVcfRecordComparator`. When the list of records to be sorted contains chromosome values that aren't in the list of "reference" contigs, those records sort to the top because their indexOf is -1. Further, those records are then sorted within themselves only by position because of that's the test in the `if (diff == 0)` block. As a result, the list `[(chr1, 100), (chr1, 200), (chr2, 400), (chr3, 300)]` sorted against a comparator created with just one reference contig `[chr1]` has the ordering: `[(chr3, 300), (chr2, 400), (chr1, 100), (chr1, 200)]`.
What happened for my case was the input bam had just one sequence line line `@SQ     SN:chr1` so when the snps from the `qsignature_positions.txt` input were sorted:
https://github.com/AdamaJava/adamajava/blob/d1983fc02f2d0c6fa511f1d56106e96e50a090be/qsignature/src/org/qcmg/sig/SignatureGeneratorBespoke.java#L190
every SNP that _wasnt_ on chr1 went to the top and the logic for how to advance through the snp records in sync with the bam is no longer based on good assumptions... specifically `match` ends up calling itself fruitlessly at line 527 after the `advanceVCFAndPosition` until the stack overflows because the chr of the first bam record doesn't match chr of any of the (first million or so) of the sorted snp records.

delocalizer:
[...]

PR #49, opened 2019-02-02
Fix for #48

delocalizer:
Implement the Comparator returned by ChrPositionComparator.getVCFRecordComparator so that it orders all records with a CHROM that appears in the list to be before all records whose CHROM value does not appear in the list


delocalizer:
I believe this is failing because after merging master back into this branch (which I needn't have done, really) the code checker in the TC build is picking up a bunch of new style issues from the qprofiler2 PR that got merged in just before.
@holmeso is there a way to override the build thresholds on case-by-case basis so it doesn't complain here?

holmeso:
> I believe this is failing because after merging master back into this branch (which I needn't have done, really) the code checker in the TC build is picking up a bunch of new style issues from the qprofiler2 PR that got merged in just before.
> @holmeso is there a way to override the build thresholds on case-by-case basis so it doesn't complain here?

I'll see if I can get it to check the inspection values from the production build rather than the PR build. Disabling for now so that you can get your change through...

delocalizer:
> > I believe this is failing because after merging master back into this branch (which I needn't have done, really) the code checker in the TC build is picking up a bunch of new style issues from the qprofiler2 PR that got merged in just before.
> > @holmeso is there a way to override the build thresholds on case-by-case basis so it doesn't complain here?
> 
> I'll see if I can get it to check the inspection values from the production build rather than the PR build. Disabling for now so that you can get your change through...

Thanks!

PR #51, opened 2019-02-04
add final to immutable string variable

ChristinaXu2017:


ChristinaXu2017:
minor change to avoid build complains

Issue #54, opened 2019-03-04
qannotate homopolymer mode throws unhelpful exception


holmeso:
Liam is running qannotate's homopolymer mode (as part of the annotation process rather than in isolation) and it is giving the following error:

```
18:09:35.374 [main] TOOL au.edu.qimr.qannotate.modes.HomoplymersMode - window size for homoplymers: 100
18:09:35.374 [main] TOOL au.edu.qimr.qannotate.modes.HomoplymersMode - report homoplymers base on both side of variants: 10
java.lang.ArrayIndexOutOfBoundsException: -1
        at au.edu.qimr.qannotate.modes.HomoplymersMode.findHomopolymer(HomoplymersMode.java:194)
        at au.edu.qimr.qannotate.modes.HomoplymersMode.getHomopolymerData(HomoplymersMode.java:134)
        at au.edu.qimr.qannotate.modes.HomoplymersMode.annotate(HomoplymersMode.java:95)
        at au.edu.qimr.qannotate.modes.HomoplymersMode.addAnnotation(HomoplymersMode.java:81)
        at au.edu.qimr.qannotate.modes.HomoplymersMode.<init>(HomoplymersMode.java:61)
        at au.edu.qimr.qannotate.Main.main(Main.java:42)
main java.lang.ArrayIndexOutOfBoundsException: -1 : -1
```

Would it be possible to throw an exception that has some more information to hopefully allow the user to address whatever issue has arisen?

**To Reproduce**
1. Run qannotate homopolymer mode against the following vcf:
`/working/genomeinfo/cromwell-test/cromwell-executions/somaticDnaARGSToMaf/4c9d0a08-d159-4459-abbc-d019af76eb89/call-homAnno/inputs/-1847501406/21110ac3-7403-4f2a-83c7-0d862ff86034.merged.dbsnp.germ.vcf.gz`
2. See error

Note that Liam is running this process against some modified bam files - this is not happening as part of the production workflow, and so the cause of the error may not be "standard"

holmeso:
OK, so there is a bug in `qannotate`'s `homopolymer` mode code where if there is a variant at position 1 (yeah, I know...) then an `ArrayOutOfBoundsException` is thrown.

What would biologists like to see in the HOM field when there is a variant at position 1?

holmeso:
A fix has gone in for the `ArrayOutOfBoundsException` has gone in, and results in the following entry in the `INFO` field after applying the homopolymer annotation:
```
[...]

PR #55, opened 2019-03-05
fix homopolymer bug at start/end of contigs

holmeso:
`Qannotate` was throwing an exception when running `homopolymer` mode and variants were present at the very beginning (position 1) of the contig (see #54 )

Added null guards and checks to ensure that variants at the beginnings of contigs are handled appropriately (in my view)
example output for a position at the start of a contig:
```
GL000192.1   1    .    G    C  ...   FLANK=-----CAATTC;IN=1;HOM=2,gAATTCATTCA ...
```

If a genome biologist takes a look at #54 and decides that this is not the desired output, then this will need to be revisited, but for now, it allows `qannotate` to proceed error free.

Have also added some checks for variants that are beyond the ends of the contigs.

Added more meaningful error messages when options are missing from `qannotate`.

### DIFF
--- a/qannotate/src/au/edu/qimr/qannotate/Main.java
+++ b/qannotate/src/au/edu/qimr/qannotate/Main.java
@@ -16,58 +16,105 @@ public class Main {
 	public static void main(final String[] args) throws Exception {	
 
 		try {
-            final Options options = new Options(args);             
+           final Options options = new Options(args);             
         		//LoadReferencedClasses.loadClasses(Main.class);    
            logger = QLoggerFactory.getLogger(Main.class, options.getLogFileName(),  options.getLogLevel());	            		               
-           logger.logInitialExecutionStats(options.getPGName(), options.getVersion(),args);	    
-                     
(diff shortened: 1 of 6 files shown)


qannotate/src/au/edu/qimr/qannotate/Main.java
```diff
@@ -16,58 +16,105 @@
...
+		 */
+		case overlap:
+		case confidence:
+		case indelconfidence:
+		case ccm:
+		case make_valid:
+			if (null == o.getOutputFileName()) {
+				throw new IllegalArgumentException("Please supply an output file using the \"-output\" or \"-o\" option");
+			}
+			if (null == o.getInputFileName()) {
```
delocalizer: is this check for input repeated below?

holmeso: It is, but the `vcf2maf` modes can take either `output` or `outdir` as options, whereas the other modes need `output`
I'll add some comments

Issue #18, opened 2018-11-06
Tidy up qannotate Overlap mode


holmeso:
qannotate Overlap modes needs to be production-ified

- Code needs to be modularised where possible.
- Tests need to be added.
- some key parameters need to become options (eg. number of threads used)

Katia has expressed a desire for the overlap mode to become part of the new pipeline, hence this feature request.

PR #57, opened 2019-03-14
Overlap in qsnp

holmeso:
Re-jig the way that `qsnp` looks at `SAMRecords` so that instead of an int identifier (position of read in contig), a long hashcode (using `lz4-java`) of the readname is used.

This allows `qsnp` to use the readname hashcode to determine if overlapping reads are present at a position.
If overlapping reads are present, and they contain the same base, this is counted once. If they contain different bases, the read pair is discarded.

Katia has given the green (ok, green-ish) light for this:

> I think I’m happy thanks….

Negates the need for #18 from a snp perspective. May still be required for indels...


qcommon/test/org/qcmg/common/util/PileupElementLiteUtilTest.java
```diff
@@ -186,45 +182,45 @@ public void testPassesCountCheckSecondPass() {
...
-		assertEquals(1, PileupElementLiteUtil.getLargestVariantNovelStarts(accum, 'C'));
-		byte C = 'C';
-		accum.addBase(C, (byte) 1, true, 1, 1, 2, 1);
-		accum.addBase(C, (byte) 1, true, 2, 1, 2, 1);
-		assertEquals(1, PileupElementLiteUtil.getLargestVariantNovelStarts(accum, 'C'));	// still have the A in there
-		assertEquals(2, PileupElementLiteUtil.getLargestVariantNovelStarts(accum, 'A'));
-		assertEquals(2, PileupElementLiteUtil.getLargestVariantNovelStarts(accum, 'G'));
-		assertEquals(2, PileupElementLiteUtil.getLargestVariantNovelStarts(accum, 'T'));
-	}
+//	@Test
```
delocalizer: I see there are several commented-out tests in this ci; is that because they might be resurrected in some form?

PR #64, opened 2019-05-07
seperate common code into old qprofiler and qprofiler2

ChristinaXu2017:


ChristinaXu2017:
not yet ready. withdraw this action

PR #65, opened 2019-05-08
Qprofiler may2019

ChristinaXu2017:
try to merge updated qprofier2 into master, it will be very different from the previous one which was on the master.  next step will separate qprofiler related code from qcommon to qprofiler1 or qprofiler2


ChristinaXu2017:
build error and I  will try again

PR #70, opened 2019-05-22
Qprofiler may2019

ChristinaXu2017:
remove artifacts from build file, It make fatJar task independent, unless run "gradle fatJar" and then it will create qprofiler2-all-2.0.jar; 
remove "/" from comment out line from xsd file

### DIFF
--- a/qprofiler2/build.gradle
+++ b/qprofiler2/build.gradle
@@ -34,4 +34,6 @@ task fatJar(type: Jar ) {
 		exclude 'META-INF/.RSA', 'META-INF/.SF','META-INF/*.DSA'		
 }
 
-//artifacts { archives fatJar }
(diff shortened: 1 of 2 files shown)


ChristinaXu2017:
the artifacts { archives fatJar } is commented out at the moment, but not sure to who and when the fatJar is required, we need further discussion and investigation. let it stay there as note

holmeso:
> the artifacts { archives fatJar } is commented out at the moment, but not sure to who and when the fatJar is required, we need further discussion and investigation. let it stay there as note

If you don't know if the fatJar is required, why are you commenting it out? Would it not make more sense to either leave it in, or to delete it?

ChristinaXu2017:
fixed

PR #72, opened 2019-05-30
bug fix for bams with lots of clipping

### LINKED ISSUE #71: QSV out of memory when run against BAM with lots of clipping
QSV runs out of memory when running against a BAM that has lots of clipped reads.

**To Reproduce**
1. Run QSV against a BAM with lots of clipped reads
2. Wait for it to fail

**Expected behavior**
QSV should be able to handle BAMs with lots of clipping.

**Additional context**
A BGI PANDA tumour BAM is what caused this issue to come to light.
It is most likely that the BAM itself is compromised, but QSV should still be able to handle this (I think)

Related??: http://kace01s.adqimr.ad.lan/userui/ticket.php?ID=77880

holmeso:
fixes #71

Updated `AnnotateFilterMT.addToPairWriter` so that it passes the count to `add2queue` rather than the start position of the record.
This means that every 100000 records, the size of the queue is checked, and if its larger than 500000, it will sleep until less than 500000 before resuming its work.

Also changed the number of filtering threads from 3 to 2. When running in cancer mode, 2 instances of `AnnotateFilterMT` are created, one for the control bam and one for the test bam. This class was creating 6 threads (1 of which was albeit very short-lived), 3 to filter, and 2 to write output (if running in both clip and pair mode, which I believe is the default).
And so, if this process is started with 8 threads overall, the annotate step itself is using 10. By changing the number of filtering threads from 3 to 2, this should come back to 8.

Following on from that, I've also removed the short lived thread that was populating a queue. The queue is now instantiated with what it needs.

Also removed some of the many calls to `SAMRecord.getReadGroup` as this was taking a lot of time (if you look at the SAMRecord source, it is performing a getAttribute call followed by a map lookup).

### DIFF
--- a/qsv/src/org/qcmg/qsv/QSVPipeline.java
+++ b/qsv/src/org/qcmg/qsv/QSVPipeline.java
@@ -72,10 +72,10 @@ public class QSVPipeline {
 	private final Map<PairGroup, Map<String, List<DiscordantPairCluster>>> normalRecords;
 	private int somaticCounts = 0;
 	private int  germlineCounts = 0;
-	private int normalGermlineCounts =0;
+	private int normalGermlineCounts = 0;
 	private final String analysisId;
 	private final String resultsDir;
-	private long clipCount =0;
+	private long clipCount = 0;
 
 	public QSVPipeline(Options options, String resultsDir, Date analysisDate, String analysisId, QExec exec) throws Exception {
 
--- a/qsv/src/org/qcmg/qsv/annotate/AnnotateFilterMT.java
+++ b/qsv/src/org/qcmg/qsv/annotate/AnnotateFilterMT.java
@@ -252,17 +252,7 @@ public class AnnotateFilterMT implements Runnable {
 					chromosomes = queueIn.poll();               
 
 					if (chromosomes == null) {
-
-						// qIn maybe filled again during sleep, so sleep should
-						// be secondly
-						try {
-							Thread.sleep(sleepUnit);
-							sleepcount++;
-						} catch (final InterruptedException e) {
-							logger.info(Thread.currentThread().getName() + " "
-									+ e.toString());
-						}
-
+						run = false;
 					} else {
 						// pass the record to filter and then to output queue                    	
 						for (final Chromosome chromosome: chromosomes) {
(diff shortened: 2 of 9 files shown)


holmeso:
Could any potential approver hold fire for the moment pls - I'm investigating an issue...

holmeso:
The bug fix brought to light another bug where the thread would prematurely die thinking that the queue it was operating on was empty. I presume that this didn't materialise when the original bug was present because as there was nothing to hold back the filtering threads, they would always have completed before the writing thread had an empty queue.

holmeso:
> Could any potential approver hold fire for the moment pls - I'm investigating an issue...

all good now

ChristinaXu2017:
just a suggestion. There were some pairs missed out inside qProfiler1 when the method "isF5toF3()" and "isF3toF5()" are called. A pair with the same start position and same orientation, will not belong to F5toF3 or F3toF5, they are missed out but they should be either F5toF3 or F3toF5. 
I updated the qprofiler2, try to keep these pairs and mark them as  F3F5. 
I am not sure whether you try to discard these pairs to keep them.

holmeso:
> just a suggestion. There were some pairs missed out inside qProfiler1 when the method "isF5toF3()" and "isF3toF5()" are called. A pair with the same start position and same orientation, will not belong to F5toF3 or F3toF5, they are missed out but they should be either F5toF3 or F3toF5.
> I updated the qprofiler2, try to keep these pairs and mark them as F3F5.
> I am not sure whether you try to discard these pairs to keep them.

OK, good to know - thanks for that!. I will check with Felicity as to what is specifically required.
Is your code with the more accurate `isF5toF3` and `isF3toF5` methods in `qcommon`?

ChristinaXu2017:
they are not inside qcommon, just inside qprofiler2. If I put them into qcommon, it will affect qprofiler1 and other tools at the moment.

delocalizer:
I intended to do this Friday, then yesterday, and now today... stuff keeps happening. If I don't get to it tomorrow I may unassign myself in defeat :frowning_face:

ChristinaXu2017:
The details of paring orientation method is defined inside 
org.qcmg.qprofiler2.summarise.PairSummary.getPairType(). It is better to merge your code to master now, you can update the paring on next merge.

PR #74, opened 2019-06-17
Qprofiler jun2019

ChristinaXu2017:
1. add attribute readCount to all <sequenceMetrics> under <bamMetrics><bam field>. Added unit tests, make sure these counts are consistence to <bamSummary>. 
2. add attribute file_checksum to <bamReport>, so NSG check can know whether two input BAM with same name are identical or not.

### DIFF
--- a/qprofiler2/src/org/qcmg/qprofiler2/summarise/PositionSummary.java
+++ b/qprofiler2/src/org/qcmg/qprofiler2/summarise/PositionSummary.java
@@ -13,6 +13,8 @@ package org.qcmg.qprofiler2.summarise;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
@@ -27,8 +29,7 @@ public class PositionSummary {
 	private final AtomicInteger min;
 	private final AtomicInteger max;
 	private final QCMGAtomicLongArray[] rgCoverages; // the coverage for each readgroup on that position   
-//	private final QCMGAtomicLongArray coverage = new QCMGAtomicLongArray(512);  // total coverage on that position
-	private final List<String> readGroupIds;
+	private final List<String> readGroupIds;  //nature sorted
 	
 	private final ArrayList<Long> maxRgs = new ArrayList<Long>(); //store the max coverage from all read group at each position;
 	private Boolean hasAddPosition = true; //set to true after each time addPosition, set to false after getAverage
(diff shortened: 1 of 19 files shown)


qprofiler2/src/org/qcmg/qprofiler2/util/CycleSummaryUtils.java
```diff
@@ -9,8 +9,8 @@
 import htsjdk.samtools.Cigar;
 
 public class CycleSummaryUtils {
-	public static final Pattern BAD_MD_PATTERN = Pattern.compile("([ACGTN])");	
-	private static final Pattern badReadPattern = Pattern.compile("([.N])");	
+//	public static final Pattern BAD_MD_PATTERN = Pattern.compile("([ACGTN])");
```
holmeso: delete

qprofiler2/src/org/qcmg/qprofiler2/summarise/PositionSummary.java
```diff
@@ -152,7 +156,7 @@ public void addPosition(final int position, String  rgid  ) {
 		
 		rgCoverages[ readGroupIds.indexOf(rgid)  ].increment(position / BUCKET_SIZE);
```
holmeso: what is the performance impact of looping through a list (`readGroupIds.indexOf(rgid)`) for each record?
If you don't want to use a map, would it be better to sort the array and call `binarySearch` instead?

qprofiler2/src/org/qcmg/qprofiler2/summarise/PositionSummary.java
```diff
@@ -27,7 +27,7 @@
 	private final AtomicInteger min;
 	private final AtomicInteger max;
 	private final QCMGAtomicLongArray[] rgCoverages; // the coverage for each readgroup on that position   
-	private final QCMGAtomicLongArray coverage = new QCMGAtomicLongArray(512);  // total coverage on that position
+//	private final QCMGAtomicLongArray coverage = new QCMGAtomicLongArray(512);  // total coverage on that position
```
holmeso: delete

qprofiler2/src/org/qcmg/qprofiler2/util/CycleSummaryUtils.java
ChristinaXu2017: done

qprofiler2/src/org/qcmg/qprofiler2/summarise/PositionSummary.java
ChristinaXu2017: There were only a couple readGroups in one BAM, but now sometimes we received BAM with 30 read groups. O(log(N)) is much faster than O(N), especially we only sort the list once. 
OK, the code is updated

qprofiler2/src/org/qcmg/qprofiler2/summarise/PositionSummary.java
ChristinaXu2017: done!

PR #82, opened 2019-07-18
change to getFullChromosome so that if supplied ref is chrM, that is …

holmeso:
…what will be returned rather than chrMT

This change will allow `qannotate` to be run against `GRCh38`

### DIFF
--- a/qcommon/src/org/qcmg/common/util/IndelUtils.java
+++ b/qcommon/src/org/qcmg/common/util/IndelUtils.java
@@ -193,16 +193,11 @@ public class IndelUtils {
 	public static String getFullChromosome(String ref) {
 		if (ref == null ) return null; //stop exception
 		
-		//convert chrMT first 
-		if (ref.equals("chrM") || ref.equals("M") || ref.equals("MT")) {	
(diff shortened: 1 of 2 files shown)


ChristinaXu2017:
Previously we convert "chrM" to "chrMT", one of the reason is DBSNP database use "chrMT".  But the new method treats "chrM" and "chrMT" different, it will affect at least dbsnp mode in qannotate.

holmeso:
> Previously we convert "chrM" to "chrMT", one of the reason is DBSNP database use "chrMT". But the new method treats "chrM" and "chrMT" different, it will affect at least dbsnp mode in qannotate.

but only if the reference that was used has chrM rather than chrMT. If so, that I don't think we can expect to use the GRCh37 version of dbSnp

holmeso:
A tagged release of adamajava has been made prior to this change being merged into master. It is `https://github.com/AdamaJava/adamajava/tree/v18jul2019`

If, for whatever reason, someone would like to run our code and have `chrM` changed to `chrMT`, then that is the version to use.
Technical debt, here we come!

Issue #84, opened 2019-07-18
Add entry to vcf header for 5BP


holmeso:
The `5BP` annotation is added to the format filter field for a sample if it has reads that are contributing to the alt that are at the ends of the read (within the last 5 bp).

There is currently no mention in the vcf header file of `5BP`. This needs to change.

May I suggest the following:

> ##FILTER=<ID=5BP,Description="There are reads contributing to the alt that are within the first/last 5 bps of the read">

I would suggest that the vcf header is added at `qannotate`'s `ConfidenceMode` stage, which is where the annotation is added based on the contents of the `EOR` field within the format section.

ChristinaXu2017:
good suggestion. I updated code, should I merge to master?

holmeso:
> good suggestion. I updated code, should I merge to master?

Not yet, I would suggest that we wait until an analyst has had a change to give this the thumbs up.

holmeso:
Description will be:

> There are fewer than 5 reads supporting the mutation where the mutation is in the middle of the read (ie. not 5bp from the start or end of the read)
OR there are 5 or more reads supporting the mutation where the mutation is in the middle of the read, BUT they are all on the same strand

Issue #85, opened 2019-07-19
COV vcf header should contain coverage cutoff values used


holmeso:
The `COV` vcf annotation has replaced `COVN8`, `COVN12`, and `COVT` in the latest vcf outputs.
The aim was to provide a generic way of indicating that there was a coverage issue. 
It also allows the user to adjust the coverage cutoff values (in `qannotate`) without the annotation losing its meaning.

However, in the vcf header, the description of the `COV` annotation is currently overly generic and gives the user no indication what coverage values were used.

> ##FILTER=<ID=COV,Description="Not enough coverage">

A couple of solutions as I see it:

1. add in the default values to the vcf header. It is rarely the case that these default values are overridden. If they are then `qannotate` could overwrite the vcf header entry with the updated values.

2. only add the vcf header entry when this is set (`qannotate`'s `ConfidenceMode`), using the values (default or otherwise) used to make the decision.

A potential downside to customising the header value would be when we come to merge the snp and indel vcf files, and different setting have been used for snps and indels. This would leave the user confused as to what coverage values applied to what variant type (unless this was explicitly defined in the header).

Thoughts?

PR #86, opened 2019-07-23
moved vcf header filter descriptions from qsnp to qannotate

holmeso:
Vcf header entries for filters applied by `qannotate` were being added by `qsnp` (hangover from when filters were added in `qsnp`).

This has now been rectified.
`qsnp` no longer adds vcf header entries for filters/info that it doesn't insert
`qannotate`'s `ConfidenceMode` now adds vcf header details for the filters that it applies.
`qannotate` will also put the cutoff values used into the header descriptions so that end users know what values were used.

Both `qsnp` and `qannotate` vcf outputs pass vcf-validation (`vcftools`)


holmeso:
fixes #84 #85

PR #90, opened 2019-07-26
Qannotate jul2019

ChristinaXu2017:
add option --strict2ChromosomeName" by default  (without this option) we accept ambiguous chr name, eg. mt == chrMT. With this option, the contig name between database file and input must exactly match, also case sensitive.  Some mode without database file, such as confidenceMode, vcf2mafMode will ignore this option.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,7 @@ subprojects {
   }
 
   dependencies {
-	testCompile 'junit:junit:4.12'	
+	testCompile 'junit:junit:4.10'	
   }
   checkstyleMain.onlyIf {project.hasProperty('checkstyle')}
    
--- a/build.gradle
+++ b/build.gradle
@@ -46,6 +46,10 @@ subprojects {
         html.enabled = false
     }
   }
+
+  dependencies {
+	testCompile 'junit:junit:4.10'	
+  }
   checkstyleMain.onlyIf {project.hasProperty('checkstyle')}
    
    sourceSets {
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,7 @@ subprojects {
   }
 
   dependencies {
-	testCompile 'junit:junit:4.10'	
+	testCompile 'junit:junit:4.12'	
   }
   checkstyleMain.onlyIf {project.hasProperty('checkstyle')}
    
--- a/qannotate/build.gradle
+++ b/qannotate/build.gradle
@@ -16,7 +16,8 @@ dependencies {
 	compile name: 'snpEff', version: '4.0e'	
 	compile 'com.github.samtools:htsjdk:2.14.1'
 	compile 'net.sf.jopt-simple:jopt-simple:4.6'
-	testCompile 'junit:junit:4.10'
+
+ 
 }
 
  
--- a/qannotate/src/au/edu/qimr/qannotate/Main.java
+++ b/qannotate/src/au/edu/qimr/qannotate/Main.java
@@ -16,10 +16,11 @@ public class Main {
 	public static void main(final String[] args) throws Exception {	
 
 		try {
-           final Options options = new Options(args);             
-        		//LoadReferencedClasses.loadClasses(Main.class);    
+           final Options options = new Options(args);
            logger = QLoggerFactory.getLogger(Main.class, options.getLogFileName(),  options.getLogLevel());	            		               
            logger.logInitialExecutionStats(options.getPGName(), options.getVersion(),args);
+	       logger.tool("logger file " + options.getLogFileName());
+	       logger.tool("logger level " + options.getLogLevel());
            
            checkOptions(options);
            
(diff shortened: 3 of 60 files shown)


ChristinaXu2017:
Thanks for these suggestions, I also wish there are the only couple of lines changes to support this new function.  I am happy to explain all mysterious parts: 

- Many many changes that seem unrelated to the purpose of the branch in SampleColumnTest, Vcf2mafTest Vcf2mafIndelTest, GermlineModeTest, DbSnpModeTest, CustomerConfidenceModeTest, ConfidenceModeTest, AbstractModeTest, Vcf2maf, TandemRepeatMode, SnpEffMode, MakeValidMode etc.

reason: 
This option "--strict2ChromosomeName"  has to pass to all subclasses of "qannotate", once a mode (subclass) read the input VCF file and a database file, they have to check whether they take strict/lenient chr name. I have to also test on all related test class, in case I missed something. 

- qpicard/src/org/qcmg/picard/Faidx.java - is that new?
This code was used in my unit test code to create a referrence index file rather than call external tools. So far only called by HomoplymersModeTest.java. I think it may be useful for future development. 

- SignatureUtil.addChrToVcfRecords - why is that commented out?
during unit updating, SignatureUtil complains, I found the method "addChrToVcfRecords" is no longer used by any project, it should be a deprecated method, it is better to delete it rather than update it. 

- Whitespace-only changes in VcfUtils, QLoggerFactory etc.
whitespace maybe happened during editing without purpose.



 meanwhile, there are some unused code, I just remove them, such as SignatureUtil.addChrToVcfRecords.

ChristinaXu2017:
regarding the build file changes, It should be under another separate branch.  However, each merge process takes several days to process. reviewer take leaves, busy with other projects, the build fails etc.  The master was merged twice since I check out the branch, to avoid conflicts and save time, I didn't create another branch. 
another reason is there are no big changes for build files, I just move repeated lines into root file, make sure we also use the same JUnit.  qannotate is a bit dated, now we use JUnit rule to create a temporary folder for testing data. So it also related to qannotate updates.

ChristinaXu2017:
I take the suggestion to change the name of "getNewchrName(...)", But it will take into next merge, otherwise it causes more confusing for the reviewer.  If we leave the branch stay alone for a too long time, I am sure there are more changes happen.

ChristinaXu2017:
There are 59 files changed, it sounds many. But 25 of them are BUILD files, each file contains one line change only.  There was a line appears in all 24 files, now it moved into root BUILD file.

ChristinaXu2017:
I personally think this merge process is Impractical, however, we have to follow it as a good team member.  Hopefully, we will get used to it.

delocalizer:
There don't seem to be any new changes to review?

ChristinaXu2017:
what new changes do you expect to have?  Have you read my comments and all changes on the 59 files?  some of your suggestion makes me very confused, eg. 

> I'm inclined to think some alternate signature for ChrPosition constructors that takes a strict argument, defaulting to false (or whatever the current behaviour is) would be more efficient. 

the ChrPosition constructors are never changed, it always takes the strict argument.  Unless the new static method is called "getNewchrName(...)". 

> getNewchrName is very misleading; it's returning a ChrPosition not a new chromosome name. The original cloneWithNewChromosomeName was perfect.

I agree "getNewchrName(...)" isn't a good name, but it should NOT be "cloneWithNewChromosomeName" as well. the old method "cloneWithNewChromosomeName()", replace the CHR name by a supplied argument "a new CHR name". our new method repace the CHR name based on the whether it should be "strict" or "lenient".  It may be "cloneIfChromosomeNameStrict(...)". it is better to leave it to the next merge, you are also welcome to change it on next merge and I am happy to review it.

ChristinaXu2017:
I guess you may suggest not to make so many changes, keep use the method "loadVcfRecordsFromFile(File f) ",  it takes lenient CHR name by default. 

I can't accept this suggestion. If we use the default method,  that means we allow them to neglect the new option "--strict2ChromosomeName".  Each mode in Qannotate has to check all options from the command line.  If this option is not used in the command line, then  a default value will pass to the  new method loadVcfRecordsFromFile(File f, boolean isStrictName) 

I also added lots of unit test code for each Qannotate mode, to make sure they can pick up this option correctly.  eg. Vcf2maf must use strict CHR name since there is no database file are loaded, they are not required to match the CHR name.

delocalizer:
I'm not suggesting anything about `loadVcfRecordsFromFile`, so there's not a lot of point starting some argument about that. The specific changes I wanted were (I thought) perfectly clearly expressed in my first comment, which were carefully considered and made after spending over an hour reviewing this PR. If you're not willing to make any changes based on the original comment then I think we're in fundamental disagreement about the purpose of code review and I will ask to be unassigned as reviewer by one of the repo admins.

holmeso:
In light of the released PR #97 , could you please review this PR and either close and re-submit, or clean up so that only the relevant changes remain

ChristinaXu2017:
pause at the moment due to the conflicts with mater branch

PR #97, opened 2019-08-02
Develop.master

ChristinaXu2017:
a. tidy up build files, adjust the space and remove JUnit test from subproject to root
b. solve Junit problem cause by snpEff-4.0e: 
    During our test compilerjunit4.10 is run, which support TemporaryFolder rule.  However, our dependency snpEff-4.0e.jar complied with Junit4.4, which disabled our junit4.10 and make TemporaryFolder rule stop working properly.  The reason is TemporaryFolder rule requires BlockJUnit4ClassRunner but junit4.4 call JUnit4ClassRunner which is old. 
    To solve the problem, we can either add a customized runner which inherits all feature from BlockJUnit4ClassRunner, or specify the compiler and runtime classpath on the build file


holmeso:
Could you please document in the PR as to why you are adding a Runner class

holmeso:
Could you also please expand on the issues around 4.10 and 4.12 wrt snpeff - thx!

holmeso:
Another way of fixing the issue so that is no need for customised Runner code is as follows:

```
configurations {
    junit
}
sourceSets.test.compileClasspath = configurations.junit + sourceSets.test.compileClasspath
sourceSets.test.runtimeClasspath = configurations.junit + sourceSets.test.runtimeClasspath
dependencies {
    junit ('junit:junit:4.10')
   ...
```

This has the effect of putting the specified version of `junit` ahead of the other jar files in the classpath, which means we don't pick up an embedded version from another jar

ChristinaXu2017:
Ok, I agree it is a better way to solve the junt test issue caused by snpEff. however, we still should put docs somewhere to address the problem and the why we do so. Should we add information into the build file?

qannotate/test/au/edu/qimr/qannotate/CallBlockJUnit4ClassRunner.java
```diff
@@ -0,0 +1,22 @@
...
+ * The reason is TemporaryFolder rule requires BlockJUnit4ClassRunner but junit4.4 call JUnit4ClassRunner which is old. 
+ * 
+ * To solve the problem we define an customized runner which inherits all feature from BlockJUnit4ClassRunner. It also added into our test class: "@RunWith(CallBlockJUnit4ClassRunner.class"
+ * Hence the BlockJUnit4ClassRunner will run again for qannotate unit test compiler. 
+ * 
+ * @author christix
+ *
+ */
+ 
+public class CallBlockJUnit4ClassRunner extends BlockJUnit4ClassRunner{
```
holmeso: Is this class still required?

holmeso:
Adding some comments to the build file to explain what is going on here is a great idea

qannotate/test/au/edu/qimr/qannotate/CallBlockJUnit4ClassRunner.java
ChristinaXu2017: leave it at the moment, in case it will be used late on.

ChristinaXu2017:
I don't think the build file is a good place to add comments. If we have to explain a reason (add comments) for this change, then every line in build file should have comments as well, they also have reason to exist in the file. This will make build file very mess.

qannotate/test/au/edu/qimr/qannotate/CallBlockJUnit4ClassRunner.java
holmeso: I suggest you take it out and keep it in a github issue in case we need it in the future. I don't want to pollute our code base with code that is not used.

holmeso:
> I don't think the build file is a good place to add comments. If we have to explain a reason (add comments) for this change, then every line in build file should have comments as well, they also have reason to exist in the file. This will make build file very mess.

Where do you suggest adding comments?

ChristinaXu2017:
OK, I add the reason to qannotate/buil.gradle above the   compile name: 'snpEff', version: '4.0e'
but not root build file

holmeso:
> OK, I add the reason to qannotate/buil.gradle above the compile name: 'snpEff', version: '4.0e'
> but not root build file

perfect

ChristinaXu2017:
Hi Ollie, could you please review again, the teamcity was complain, I have to make some changes according to the code inspection guide.

PR #99, opened 2019-08-02
update qannotate unit test without update sorceCode under src floder

ChristinaXu2017:
this branch is copied from the master, there is no new feature of qannotate. Here we just changed the exsiting test code to cope with junit 4.12


holmeso:
In light of the released PR #97 , could you please review this PR and either close and re-submit, or clean up so that only the relevant changes remain

ChristinaXu2017:
this branch is dated and no longer needed.  The customized runner is deleted, so all related changes should delete as well

PR #109, opened 2019-08-16
removed dependancy upon java 1.7

ChristinaXu2017:
# Description

The qmaftools required java1.7, actually, it only requires the javax library which is removed from java9 and but still in java8. To make our java version consistency, at the moment we stick to java8.
I also reported a related issue: https://github.com/AdamaJava/adamajava/issues/110, for our next task.  

Fixes # (issue)
1. Remove  "sourceCompatibility = 1.7" from qmaftools/buid.gradle

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)
 
# How Has This Been Tested?
I run on my local machine, it works

# Checklist:

- [ x] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ x] I have made corresponding changes to the documentation
- [ x] My changes generate no new warnings
- [ x] New and existing unit tests pass locally with my changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 The AdamaJava project holds code for variant callers and pipeline tools related to next-generation sequencing (NGS).  The code was created by members of the Queensland Centre for Medical Genomics ([QCMG](http://www.qcmg.org)) as part of their participation in the International Cancer Genome Consortium ([ICGC](http://www.icgc.org)).  The code is now being developed and maintained by the [Genome Informatics Group](http://www.qimrberghofer.edu.au/lab/genome-informatics/) at the [QIMR Berghofer Medical Research Institute](http://www.qimrberghofer.edu.au).
 
-While the repository contains code for dozens of utilities, the list below shows tools that are considered robust enough to be released for general use.  Most are written in Java and require version 1.7 although some now require version 1.8 or above.  Some tools and helper scripts may be written in other languages, primarily perl and groovy.  Some tools have dependencies on other software.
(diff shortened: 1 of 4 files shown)


holmeso:
I think removing our dependance on java 1.7 is a great idea. I don't think there is a need to specify the javax jar as it should be included in java 1.8.
Could you please update this PR to have a more appropriate title and description (eg. removed dependance on java 1.7)? Thanks!

ChristinaXu2017:
java8 was a very stable version but nowadays it seems a bit old.  Java SE 8 is due to go out of public support in early 2019. Definitely, we should remove the dependency java1.7; we should also change the required java version from java1.8 to java1.8+.

ChristinaXu2017:
There is no more "javax.xml" in jdk9+, it is better to download them from maven. Some our old projects still require it.

ChristinaXu2017:
after download "javax", our tools can even run/compile on java8+, including java 12

holmeso:
I would suggest you keep this PR about removing the 1.7 dependancy. If you could also update the main documentation page to reflect this, that would be great!
We will no doubt embark upon a java version upgrade at some point in the future, where your other concerns will be addressed.
Thanks

holmeso:
oh, and could you please update the description to reflect this - thanks again

ChristinaXu2017:
ok, just removed java1.7 on this PR, put my concerns on https://github.com/AdamaJava/adamajava/issues/110

holmeso:
Could you please also update the top level README.md file to remove occurrences of 1.7 - thx

PR #112, opened 2019-08-19
Here an option “—lenient” to qannotate

### LINKED ISSUE #83: Deprecate/delete method that modifies chromosome name
`IndelUtils.getFullChromosome` takes a chromosome name and returns a modified name in certain circumstances.
Namely, if the chromosome is a number, X, Y, M, or MT it will return that name preceded with `chr`.

The method has recently been modified (see PR #82) so that if the supplied chromosome name is `chrM` then that is what is returned rather than `chrMT`

This method was originally put in place due to the differences in naming chromosomes in different versions of the human reference genome (namely GRCh37 and hg19).

Consistent naming of chromosomes was useful as many of the databases (dbSNP, snpEff etc) used by `qannotate` would expect chromosomes to start with `chr` (apart from the GL's).

I am now of the opinion that `qannotate` should not modify chromosome names, but rather fail if it is asked to run against a database that does not contain the chromosome names that the vcf file has. This would hopefully force the user to use the correct database for the analysis.

Thoughts?

ChristinaXu2017:
# Description

Qannotate was updated (see PR #82) ,  the reason is also reported on issue #83.  qannotate will fail if the chromosome name are different between input file and database file. However it is still a feature to automatically match chromosome name, eg. 1, 2…X, Y, M, or MT  to chr1, chr2… chrX, chrY and chrM. 

Here an option “—lenient” is added, without this option, our qannotate will work on stringent way like PR#82. With this option, our qannatate will behave like before (before PR #82), and it also accept case insensitive name.  We strong believe the chromosome name should not change from input file and output file, all the name conversion only happened inside annotation process, the chromosome name in output file will exactly match the input one. 

Fixes #83 

## Type of change

At moment all changes only made inside qannotate, soon future, some of them may move to qcommon.
- [ X ] New feature (non-breaking change which adds functionality)
- [ X] This change requires a documentation update
  a. Main.java and Options.java are changed. New option “—lenient” is added. 
  b. AbstractMode.java is changed:
   b.1 loadVcfRecordsFromFile(File f, boolean isStringent) updated, it now take a new parameter. This method is called by it’s children classes, so there are another 6 files changes.
   b.2 new method “cloneIfLenient(ChrPosition cp, boolean isStringent )”, so far this method only used in qannotate. 
   b.3 new method “getFullChromosome(String ref) ”, so far this method only used in qannotate. 
 c. All *.java under test are updated. 
 d. Tidy up old comments.
 e. Used TemporaryFolder rule to replace old test IO methods. 
 f. add Fidex.java under test, which is only used for the unit test to create a referrence dictionary and index file. 

# How Has This Been Tested?
 All related unit tests are updated, all new feature, new method are tested. 

# Checklist:

- [ X] My code follows the style guidelines of this project
- [ X] I have performed a self-review of my own code
- [ X] My changes generate no new warnings
- [ X] I have added tests that prove my fix is effective or that my feature works


holmeso:
I think a better approach would be to have a mode that converts the chromosome names into the database compatible format.
The resultant vcf file would then be able to pass through the annotate process.
Another mode to revert to the original names could also be added that would be run at the end of the process.

This would vastly reduce on the 30 files that have been modified to accomodate the change in its current form.
I also don't think that this feature will be used often, and to pollute our code base so widely for something that will not be used frequently seems unwise.

ChristinaXu2017:
That is fine, just leave it on the branch, until one day some users require it. 
Also please provide more details about "seems unwise", or just means there are too many changes?   I thought we should keep updating to make tools functional and easy use. But it seems our purpose is to make fewer changes and updates, the most important is to make code reviewer happier.

PR #117, opened 2019-08-23
bump await times for threads

### LINKED ISSUE #116: Push out thread wait times
QProfiler failed on a recent run against a large BAM file (hello NIH) due to the following code:

```
// wait for threads to complete
try {
    logger.info("waiting for Producer thread to finish (max wait will be 20 hours)");
    if ( ! pLatch.await(20, TimeUnit.HOURS)) {
        // we've hit the 20 hour limit - shutdown the threads and throw an exception
        producerThreads.shutdownNow();
        consumerThreads.shutdownNow();
        throw new Exception("Producer thread has timed out");
  }
  logger.info("producer thread finished, queue size: " + (q == null ? getQueueSize(queues) : getQueueSize(q) ) );
```

Yes, it ran for more than 20 hours, and then the main thread killed everything.
From the log file:
```
17:38:51.334 [pool-1-thread-1] SEVERE org.qcmg.qprofiler.QProfiler - Exception caught whilst running summarizer for file: /working/genomeinfo/cromwell/cromwell-executions/somaticDnaFastqToMaf/c2d46703-bce5-43cd-9a9e-65d6047fe427/call-testSampleQp/inputs/1800118964/7fde7939-18a9-4e03-b6c4-bb9fad5090f1.bam
java.lang.Exception: Producer thread has timed out
        org.qcmg.qprofiler.bam.BamSummarizerMT.summarize(BamSummarizerMT.java:174)
        org.qcmg.qprofiler.QProfiler$1.run(QProfiler.java:197)
        java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        java.lang.Thread.run(Thread.java:748)
```

We are dealing more and more with large BAM files, and so I think it would be a good idea to push out the `await` times on all our threaded code.
This will require a search through the codebase to find all instances of `await`.
I would suggest doubling the times and see how long before we are back in there doing the same again...

**To Reproduce**
Run `qprofiler` with a single producer, and 3 consumers against a large bam file

**Expected behavior**
It should work.

holmeso:
# Description

Doubled the time that a thread can run for before being killed (from 20 hours to 40 hours).
This change has been made to multiple projects.
Only instances where the `await` time was less than 40 hours has this change been made (eg. there are a couple of instances were the `await` time has been set to over 100 hours. I did not double these times)

Fixes #116 

## Type of change

# How Has This Been Tested?

Existing unit test have been run. No new tests have been introduced.

# Checklist:

### DIFF
--- a/qannotate/src/au/edu/qimr/qannotate/modes/OverlapMode.java
+++ b/qannotate/src/au/edu/qimr/qannotate/modes/OverlapMode.java
@@ -108,7 +108,7 @@ public class OverlapMode extends AbstractMode{
 		// wait for threads to complete
 		try {
 			logger.info("waiting for  threads to finish (max wait will be 40 hours)");
-			pileupThreads.awaitTermination(100, TimeUnit.HOURS);
+			pileupThreads.awaitTermination(Constants.EXECUTOR_SERVICE_AWAIT_TERMINATION, TimeUnit.HOURS);
 			logger.info("All threads finished");			
 		} catch (Exception e) {
 			logger.error("Exception caught whilst waiting for threads to finish: " + e.getMessage(), e);
--- a/qannotate/src/au/edu/qimr/qannotate/modes/OverlapMode.java
+++ b/qannotate/src/au/edu/qimr/qannotate/modes/OverlapMode.java
@@ -108,7 +108,7 @@ public class OverlapMode extends AbstractMode{
 		// wait for threads to complete
 		try {
 			logger.info("waiting for  threads to finish (max wait will be 40 hours)");
-			pileupThreads.awaitTermination(40, TimeUnit.HOURS);
+			pileupThreads.awaitTermination(100, TimeUnit.HOURS);
 			logger.info("All threads finished");			
 		} catch (Exception e) {
 			logger.error("Exception caught whilst waiting for threads to finish: " + e.getMessage(), e);
--- a/qannotate/src/au/edu/qimr/qannotate/modes/OverlapMode.java
+++ b/qannotate/src/au/edu/qimr/qannotate/modes/OverlapMode.java
@@ -107,8 +107,8 @@ public class OverlapMode extends AbstractMode{
     	
 		// wait for threads to complete
 		try {
-			logger.info("waiting for  threads to finish (max wait will be 20 hours)");
-			pileupThreads.awaitTermination(20, TimeUnit.HOURS);
+			logger.info("waiting for  threads to finish (max wait will be 40 hours)");
+			pileupThreads.awaitTermination(40, TimeUnit.HOURS);
 			logger.info("All threads finished");			
 		} catch (Exception e) {
 			logger.error("Exception caught whilst waiting for threads to finish: " + e.getMessage(), e);
--- a/qbamfilter/src/org/qcmg/qbamfilter/query/QueryMT.java
+++ b/qbamfilter/src/org/qcmg/qbamfilter/query/QueryMT.java
@@ -24,6 +24,7 @@ import htsjdk.samtools.SAMRecord;
 import htsjdk.samtools.ValidationStringency;
 
 import org.qcmg.common.log.QLogger;
+import org.qcmg.common.util.Constants;
 import org.qcmg.picard.SAMFileReaderFactory;
 import org.qcmg.picard.SAMOrBAMWriterFactory;
 import org.qcmg.picard.SAMRecordFilterWrapper;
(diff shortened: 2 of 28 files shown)


ChristinaXu2017:
1. I am confused, why there are 16 files changed. they seem not related to this topic
2. I strongly believe we should not increase await time from 20 to 40.  The bottleneck in 7fde7939-18a9-4e03-b6c4-bb9fad5090f1.bam.qp.log is it sleep too much during loading reads to queue. That means it requires more consumer to process these queued reads. Even we double the await time, it will not solve the bottleneck.

ChristinaXu2017:
I run this BAM on qprofiler2 with -ntP 3 -ntC 20. It seems 2-3 times fast than the failed one. I will try it on old qrofiler with a single producer and 10 consumers.

holmeso:
The time taken to run through a bam increases as the size of the BAM increases. We are now dealing with much larger BAM files than we have dealt with in the past (I think the original time limit of 20 hours has been around for many years).

It won't always be possible to throw extra resources at a program (eg. increase the number of consumer threads) as the user may be constrained by hardware etc.

And so, it makes sense to increase (double) the time that we allow a thread to operate upon a BAM before killing it.

> not sure why so many files are changed

I've made this change for all of our tools in `adamajava` that had the 20 hour internal "walltime" set. Some of these tools may be inflexible in the number of threads that are used which means that increasing this internal "walltime" would be the only way to prevent them from timing out.

If there are any files in particular that you would like to be excluded from this, please could you highlight them for further discussion? Thanks!

delocalizer:
@ChristinaXu2017 The 16 files changed are all directly related to this topic, as I was able to observe by reading the description and then looking at the changes. This idiom of waiting for 20 hours (or whatever) has been used throughout the codebase.
You are correct that it is possible to avoid a specific qprofiler failure by re-running with increased consumer/producer threads (as discussed in Slack, this is indeed the current workaround) but the general issue has to be seen in the context of our operations processes, i.e. running our automated cromwell workflows efficiently on shared HPC resources, which Ross has been successfully accomplishing for the last 12 months for a variety of different datasets.
Currently qprofiler tasks request 4 cores and that's fine for 99% of cases; if we bump that up across the board in the WDL to 16 or more cores just so the odd large bam we hit 1% of the time escapes the 20hr thread timeout then 99% of the time we're requesting more cores than we actually need to finish in reasonable time. And we have the same issue with all the other multithreaded tools: for a given number of threads, an unusually large bam may cause them to die simply because of some arbitrary internal timeout limit. So short of some solution involving dynamic determination of ncpu at runtime based on input filesize, the fix as implemented in this PR is both simple (total 36 lines of code) and results in better usage of compute resources than simply upping resource request defaults in the WDL.
It'd be great if you would approve this PR as-is so we can push on with a new release.

ChristinaXu2017:
If a user doesn't have enough resources (eg. limited cups on a laptop),  It is impossible to process a such big BAM.

grendeloz:
My thinking is that we should be engineering for the main use case for our tools - whole genome tumour/normal analysis on a cluster with decent size nodes. Working on laptops would be nice but that can't drive our thinking. And I agree that we can't keep dialling resource requests to cope with BAM size so we need to accept that there many be longer run times and make code changes accordingly.

holmeso:
OK, so I think we all agree that this code is designed to run on clusters, and so it makes sense to set the amount of time we are prepared to wait for a job to finish using the `walltime` attribute in a `qsub` command rather than in internal code.
And so, where we were calling `await(timeout)` on a `CountDownLatch`, I've replaced this with the no argument method which will wait indefinitely for the latch to count down, and in instances were we call `awaitTermination` on an `ExecutionService`, I've increased the timeout to 100 hours, in the hope that any `walltimes` being set would be less than this.
If people think I need to increase this value (100 hours) to something more infinite, please let me know.

holmeso:
> I agree to set the internal wall time infinitely (such as 100) but let the "qsub" to make the final decision on termination time. Here is a couple of question/suggestion:
> 
> 1. a bit confused, why 100 hours and 20 hours mixed in class:
>    qbasepileup/src/org/qcmg/qbasepileup/indel/IndelBasePileupByChrMT.java.
> 2. I saw the 100 hours appears many classes, how about to set a constant on qcommon. It will make future code maintains easy.

Constant value has been added to `qcommon`'s `Constants` and is now used.
I had set some wait times to 20, when they appeared after a 100 hour wait time. I have set them all to 100 now to avoid confusion.

PR #121, opened 2019-08-30
new Compare class that performs comparison regardless of underlying v…

### LINKED ISSUE #120: Single compare class for qsignature
`QSignature` can produce 2 different types of vcf file. There is the "traditional" vcf output, which has a line for each entry in the qsignature snp positions file, regardless of whether there is coverage at that position.
There is also the "bespoke" vcf file, which only outputs an entry if there is coverage at the position. It also outputs coverage for each readgroup in the bam as well as a total. This allows `qsignature` to be run against merged bams and the ability to detect mix-ups within a merged bam exists.

Some of our studies have been run using the traditional vcf output, and some with the newer (and recommended) bespoke vcf output.

Unfortunately, there is not a single `qsignature` comparison tool to compare the traditional vcf file with the bespoke vcf file.

It would be great it there was a single `Compare` class that would compare traditional and bespoke vcf files.
This is now particularly pressing as there is a requirement to have the `qsignature` comparison run on a nightly bases against all studies.

An alternative would be to know what type of files you are dealing with before hand, and then using the appropriate compare tool. This alternative would be impractical given the large number of studies we have.
Another option would be to re-do all `qsignature` vcf files so that they are in the "bespoke" format. This approach would take a large amount of resource, and require a significant amount of data to be brought back from tape, as it is typically the older projects that have been generated using the traditional vcf format.

**Additional context**
This will allow the existing `CompareBespokeGenotype` and `SignatureCompareRelatedSimpleGenotype` classes to be deprecated.

holmeso:
…cf type (traditional or bespoke)

# Description

- Addition of a new `Compare` `qsignature` class that is vcf agnostic.

This class will handle the comparisons for both "traditional" and "bespoke" vcf files

- Deprecation of existing classes to compare the "traditional" and "bespoke" `qsignature` vcf files.

Fixes #120 

## Type of change

- [x] This change requires a documentation update

# How Has This Been Tested?

Additional unit tests have been written to check that the Compare class handles traditional vs traditional, bespoke vs bespoke, and traditional vs bespoke.
I have also checked that the results produced by the new Compare tool produce the same output as the existing tools.

# Checklist:

### DIFF
--- a/qsignature/src/org/qcmg/sig/Compare.java
+++ b/qsignature/src/org/qcmg/sig/Compare.java
@@ -6,72 +6,70 @@
  */
 package org.qcmg.sig;
 
-import gnu.trove.map.hash.THashMap;
-import gnu.trove.map.hash.TIntShortHashMap;
-
 import java.io.File;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
-
+import org.apache.commons.math3.util.Pair;
 import org.qcmg.common.log.QLogger;
 import org.qcmg.common.log.QLoggerFactory;
-import org.qcmg.common.util.DonorUtils;
+import org.qcmg.common.util.Constants;
 import org.qcmg.common.util.FileUtils;
 import org.qcmg.common.util.LoadReferencedClasses;
 import org.qcmg.sig.model.Comparison;
+import org.qcmg.sig.model.SigMeta;
 import org.qcmg.sig.util.ComparisonUtil;
 import org.qcmg.sig.util.SignatureUtil;
+
+import gnu.trove.map.TMap;
+import gnu.trove.map.hash.THashMap;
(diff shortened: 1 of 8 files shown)


holmeso:
> Not sure what it is doing.
> It seems this Compare class will replace the previous CompareGenotypeBespoke and SignatureCompareRelatedSimpleGenotype. So it sounds a good plan and it is also great to have related unit test updated.
> However, there are still several other compare class exists, such as CompareRGGenotype, CompareIlluminaData, QsigCompare. I just doubt the name of "Compare", whether it is the main class or basic/central one for other classes partially named "compare".

Thank you for your comments. I agree that the name `Compare` is a little contentious, but the reason that this was chosen was to avoid having names like `SignatureCompareRelatedSimpleGenotype` which, albeit descriptive, are a little unwieldy.

It is my goal to eventually have the `Compare` class handle all the different types of comparison, but that is beyond the scope of this particular PR.

The scope of this PR is to have a single class that handles both the "traditional" and "bespoke"  vcf formats, which is what the new `Compare` class does.

I will create a new issue relating to the new `Compare` class integrating the remaining `*Compare*` comparisons so that this work is not forgotten about, and handled in its own PR.

PR #132, opened 2019-10-09
allow zero child for variableGroup

ChristinaXu2017:
# Description

Sometimes there are no reads from one of readGroup mapped to a contig. We have to allow zero child element under tag \<variableGroup\>. 

Fixes # (issue)

## Type of change
minor changes, just adjust one parameter in XSD file

# How Has This Been Tested?
There was an xml can't be validated by previous XSD file, but is passed by this updated XSD file. 
 

# Checklist:


holmeso:
Do you have any tests that check the bounds of these elements?

ChristinaXu2017:
Sorry for missing information on the Description page, it is updated. I did test it as shown on the page.

PR #133, opened 2019-10-23
Qanno min

### LINKED ISSUE #47: Query the values for MIN threshold in qsnp
In `org.qcmg.snp.util.GenotypeUtil` we have:
`MUTATION_IN_NORMAL_MIN_COVERAGE = 3`
`MUTATION_IN_NORMAL_MIN_PERCENTAGE = 5`
The min percentage in particular seems too low in practice, since for 20x coverage (not uncommon for a control sample) this means that a sequencing error on just one read in the normal will trigger the MIN annotation for an otherwise ok SNP call at that location in the test sample, and the call will fail filtering.
That this isn't just a theoretical problem can be seen by using qasim to generate mutations on simulated HiSeq2500 reads. For the example in https://genomeinfo.qimrberghofer.edu.au/svn/genomeinfo/adhoc/conradL/work_record/20181213_wf_evaluation, there are 14 false negative somatic calls (true mutations, uncalled) on chr1, three of them are because of this MIN annotation and all of them caused by a sequencing error on a single read.
It should also be considered that the issue will be more severe for noisier sequencing tech e.g. XTen.
Even from a biological p.o.v. it's hard to justify a 5% threshold for a normal control when you'd expect something close to either 50% or 100%. There would certainly be a case for a low threshold when the control sample was not normal, e.g. precancerous lesion or other clonal tissue from the Barrett's study, but I imagine that should be the unusual case, not the default. Perhaps this should be a tunable parameter in the qsnp configuration, with a reasonably high default e.g. 20% but settable to something lower for special cases?

holmeso:
# Description

Have changed the default values that dictate when the `MIN` annotation is added to a vcf record.
Also changed is the manner in which the default values are used.
Previously, If a hard cutoff and percentage value were present, the `MIN` annotation would be added if either the hard cutoff OR the percentage value were breached by the number of reads containing the mutation in the normal (control).
eg.
` if (altCountInControl > hardCutoff || altCountInControl > percentageCutoff)`

What happens now is that the maximum value of the cutoff and percentage is used.
eg. 
`if (altCountInControl > MAX(hardCutoff, percentageCutoff)`

The default values are now: `hardCutoff = 2 (down from Integer.MAX_VALUE)`, `percentageCutoff = 3 (unchanged)`.

This means that for coverages less than 66, 2 or more reads in the control must support the mutation before the `MIN` annotation is added.

Fixes #47 

# How Has This Been Tested?

Additional tests have been added to check the behaviour of the `MIN` annotation.
I have also run the updated code against an existing vcf file, and eye-balled the differences to ensure that vcf records that no longer contained the MIN annotation did so because of the change in cutoff value.

# Checklist:

### DIFF
--- a/qcommon/src/org/qcmg/common/vcf/VcfUtils.java
+++ b/qcommon/src/org/qcmg/common/vcf/VcfUtils.java
@@ -1504,22 +1504,31 @@ public class VcfUtils {
 	 * @param minCoverage
 	 * @return
 	 */
-	public static boolean mutationInNormal(int altCount, int totalReadCount, float percentage, int maxCoverage) {
-		if (altCount == 0) {
+	public static boolean mutationInNormal(int altCount, int totalReadCount, float percentage, int minValue) {
+		if (altCount == 0 || altCount < minValue) {
 			return false;
 		}
 		/*
-		 * if altCount is greater than the maxCoverage
+		 * Need to get the max(minValue, percentage * totalReadCount) and see if the alt count is larger than that
 		 */
-		if (altCount >= maxCoverage) {
-			return true;
-		}
-		/*
-		 * calculate percentage
-		 */
-		float passingCount = totalReadCount > 0 ? (((float)totalReadCount / 100) * percentage) : 0;
-		return altCount >= passingCount;
+		return altCount >= Math.max(minValue, (totalReadCount > 0 ? (((float)totalReadCount / 100) * percentage) : 0));
 	}
+//	public static boolean mutationInNormal(int altCount, int totalReadCount, float percentage, int maxCoverage) {
+//		if (altCount == 0) {
+//			return false;
+//		}
+//		/*
+//		 * if altCount is greater than the maxCoverage
+//		 */
+//		if (altCount >= maxCoverage) {
+//			return true;
+//		}
+//		/*
+//		 * calculate percentage
+//		 */
+//		float passingCount = totalReadCount > 0 ? (((float)totalReadCount / 100) * percentage) : 0;
+//		return altCount >= passingCount;
+//	}
 	public static boolean mutationInNormal(int altCount, int totalReadCount, int percentage, int maxCoverage) {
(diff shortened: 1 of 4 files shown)


qcommon/src/org/qcmg/common/vcf/VcfUtils.java
```diff
@@ -1504,22 +1504,31 @@ public static String getConfidence(VcfRecord rec) {
...
 		 */
-		if (altCount >= maxCoverage) {
-			return true;
-		}
-		/*
-		 * calculate percentage
-		 */
-		float passingCount = totalReadCount > 0 ? (((float)totalReadCount / 100) * percentage) : 0;
-		return altCount >= passingCount;
+		return altCount >= Math.max(minValue, (totalReadCount > 0 ? (((float)totalReadCount / 100) * percentage) : 0));
```
delocalizer: Is the check for > 0 necessary?

qcommon/src/org/qcmg/common/vcf/VcfUtils.java
```diff
@@ -1504,22 +1504,31 @@ public static String getConfidence(VcfRecord rec) {
...
+//		 */
+//		if (altCount >= maxCoverage) {
+//			return true;
+//		}
+//		/*
+//		 * calculate percentage
+//		 */
+//		float passingCount = totalReadCount > 0 ? (((float)totalReadCount / 100) * percentage) : 0;
+//		return altCount >= passingCount;
+//	}
```
delocalizer: rm unless keeping for archaeological reasons

holmeso: gone

qcommon/src/org/qcmg/common/vcf/VcfUtils.java
holmeso: no its not - gone

holmeso:
thanks Conrad - I've made changes which unfortunately means you will need to re-approve...

PR #139, opened 2019-11-12
Qsig SignatureGeneratorBespoke output options

### LINKED ISSUE #138: Qsignature output option doesn't work in GenerateBespoke mode
**Describe the bug**

The `output` runtime option when running `qsignature`'s `SignatureGeneratorBespoke` does not work.
Users must instead specify the `dir` (or `-d`) option that will set the output folder, with the name of the output vcf file coming from the name of the input file.

eg.
If `-output` `<location_and_name_of_output_file>` is specified, the output will actually go to `input_folder_and_name.qsig.vcf`
if `-dir` option is set to `<my_folder>` then output will go here: `my_folder/input_name.qsig.vcf`
If no `output` or `dir` option is specified then the output will go here: `input_folder_and_name.qsig.vcf`

This means that there is no current way of naming the qsignature output file something different to the input file.
It also means that should you wish to have multiple qsignature files for a single input file (run against different snp positions files say) then this is not possible unless you put each of them in a seperate folder.

**To Reproduce**

```
java -cp ${ADAMA_HOME}/build/lib/qsignature-1.0.jar \
   org.qcmg.sig.SignatureGeneratorBespoke \
   -snpPositions ~/qsignature_exomeid_JP_20191109.txt \
   -input ~/Devel/github/adamajava/qpileup/test/test.bam \
   -output /working/genomeinfo/users/johnPe/qsignature_exomeid/ \
   -log qsig.log -loglevel DEBUG
```

**Expected behavior**

The `output` option needs to be adhered to allowing users to tell `qsignature` what to call their outputs.

**Additional context**
This will also enable multiple `qsignature` files to exist for a given input file.

holmeso:
# Description

`Qsignature`'s `SignatureGeneratorBespoke` has been updated so that the `output` option is recognised and acted upon.

- If the user specifies an `output` that is not a directory, then the generated output will be put there.

- If the user specifies an `output` that is a directory, then the generated output will be put into that directory but will take the name of the input with the appropriate extension (`.qsig.vcf.gz`)

- If the user specifies the `dir` option then the generated output will be put into that directory but will take the name of the input with the appropriate extension (`.qsig.vcf.gz`) **NOTE** that this is the current behaviour

- If the user specifies neither the `dir` option not the `output` option, then the generated output will be live next to the input file and will have the name of the input with the appropriate extension (`.qsig.vcf.gz`) **NOTE** that this is the current behaviour

Fixes #138 

## Type of change

# How Has This Been Tested?

Additional unit tests have been added to check that this works as expected.

# Checklist:

### DIFF
--- a/qsignature/src/org/qcmg/sig/SignatureGeneratorBespoke.java
+++ b/qsignature/src/org/qcmg/sig/SignatureGeneratorBespoke.java
@@ -58,7 +58,6 @@ import org.qcmg.picard.SAMFileReaderFactory;
 import org.qcmg.picard.util.BAMFileUtils;
 import org.qcmg.picard.util.SAMUtils;
 import org.qcmg.record.Record;
-import org.qcmg.sig.model.BaseReadGroup;
 import org.qcmg.sig.util.SignatureUtil;
 import org.qcmg.tab.TabbedFileReader;
 import org.qcmg.tab.TabbedRecord;
@@ -82,6 +81,7 @@ public class SignatureGeneratorBespoke {
 	private int arraySize;
 	private int arrayPosition;
 	private String outputDirectory;
+	private String outputFile;
 	
 	private int minMappingQuality = 10;
 	private int minBaseQuality = 10;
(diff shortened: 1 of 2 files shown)


ChristinaXu2017:
It seems the updated option "--output" can handle both file and directory, why you still keep the option "--dir"?   I think the option "--dir" should be deleted, otherwise, it is quite confusing.

holmeso:
> It seems the updated option "--output" can handle both file and directory, why you still keep the option "--dir"? I think the option "--dir" should be deleted, otherwise, it is quite confusing.

I'm keeping the `dir` option in there for backward compatibility.
If in the future we want to remove that option then that would be fine.

ChristinaXu2017:
It is better to make a clear rule rather than ambiguous.  You should remove the redundant option now.

holmeso:
> It is better to make a clear rule rather than ambiguous. You should remove the redundant option now.

It is not a redundant option - it is currently used in production, hence my reluctance to remove it.

Issue #146, opened 2020-01-13
to replace "S" by "b" when output BAM FLAG in qprofiler2


ChristinaXu2017:
A pPuUrR12 table was announced In the GATK discussion section: https://gatkforums.broadinstitute.org/gatk/discussion/7019/sam-flags-down-a-boat. Our qprofiler and qprofile2 were followed this table to present BAM Flag.  However recently we found our xml output using the letter "S" to present supplementary alignment, but not "b" as shown on the pPuUrR12 table.

We plan to fix it and replace "S" with "b" for supplementary alignment in qprofiler2.

ChristinaXu2017:
fixed

Issue #144, opened 2020-01-06
QNAME update


ChristinaXu2017:
qProfiler2 check read quality first, and then report QNAME summary. However, only good reads' QNAME is summarised, which excludes discarded reads, notProperPaired reads, unmapped reads, duplicated reads.  QName often contains important information, the researcher is also interested in these bad reads QName information.

It is better to report QName information regardless of whether they are good or bad. 

I will update qprofiler2, report QNAME for all input reads as we deal with FLAG and MAPQ.

ChristinaXu2017:
Below xml have two read groups, one of them is missing inside QNAME duet to they are all bad reads
/mnt/lustre/working/genomeinfo/sample/8/2/823a6b86-fbd5-406a-9742-25debd6b95fa/aligned_read_group_set/62850d3b-942b-469d-9458-28f594c2d1a8.bam.qp2.xml

PR #147, opened 2020-01-13
Develop.master

### LINKED ISSUE #140: Handle pairing information with different tLen algorithm
The definition of TLEN in SAM specification is "signed observed Template LENgth. If all segments are mapped to the same reference, the unsigned observed template length equals the number of bases from the leftmost mapped base to the rightmost mapped base. The leftmost segment has a plus sign and the rightmost has a minus sign. The sign of segments in the middle is undefined. It is set as 0 for single-segment template or when the information is unavailable."

However different read alignment tools may have a slightly different algorithm to calculate TLEN value. For example, the bwa-meth will assign "0" to tLen if the reads marked as "not proper mapped pair" even they mapped to the same reference.  While the bwa-mem will still assign a real value to tLen, which is the " tLen = ( forwardMate.start /reverseMate.End )  - ( forwardRead.start / reverseRead.End )  +/- 1 ".   In the case of the bwa-mem reads with "tLen==0", they may be the pair with the same strand and same start/end, obviously, these pairs are overlapped. 

Current qprofile2 collects pairing information and calculates overlap bases based on the bwa-mem tLen algorithm. Several bwa-meth BAMs are failed to pass to qprofiler2. 

We should pay special attention to reads with "tLen==0". We have to check whether the pair with the same strand and the same start/end position. If not we will stop to calculate the overlapping base due to the "TLEN" value is unavailable. Else if yes, we use the current algorithm to calculate the overlapping base. 

Catch all possible exceptions rather than make the process fail. output a warning message to the log file. 

**Additional context**
couple failed run contains error message:
/genomeinfo/sample/e/6/e69526e2-499d-4288-9701-e84f6e83d1ef/aligned_read_group_set/0dfc76cb-5547-4d53-8724-77b5f7365ba7.bam.qp2.xml.log
/working/genomeinfo/sample/e/8/e87ebb22-031b-4408-8332-2863352b6094/aligned_read_group_set/7a6d0a9f-2dc7-4dcf-b8af-cca37d4a40e3.bam.qp2.xml.log

ChristinaXu2017:
# Description

1.  stop to calculate overlap base when tLen==0; 
2. use "b" rather than "S" to present supplementary reads in FLAG section;
3. report QNAME information even the reads are bad.

Fixes #140, #144 and #146

## Type of change
- [ x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

unit tests have been updated or added

# Checklist:

- [ x] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my own code
- [ x] I have commented on my code, particularly in hard-to-understand areas
- [x ] I have made corresponding changes to the documentation
- [x ] My changes generate no new warnings
- [x ] I have added tests that prove my fix is effective or that my feature works
- [x ] New and existing unit tests pass locally with my changes

### DIFF
--- a/qcommon/build.gradle
+++ b/qcommon/build.gradle
@@ -14,5 +14,7 @@ dependencies {
     compile 'org.slf4j:slf4j-api:1.7.25'
     compile 'org.slf4j:slf4j-simple:1.7.25'
     compile 'com.typesafe:config:1.3.1'
+
+    compile 'javax.xml.bind:jaxb-api:2.2.12' 
 }
 
--- a/qcommon/build.gradle
+++ b/qcommon/build.gradle
@@ -16,7 +16,7 @@ dependencies {
     compile 'com.typesafe:config:1.3.1'
     
     //jaxb is removed from JDK9+
-    //we need below jar in case to run on JDK8+
+    //we need below jar in case to run on JDK9 and above
     compile 'javax.xml.bind:jaxb-api:2.2.12' 
 }
 
--- a/qcommon/build.gradle
+++ b/qcommon/build.gradle
@@ -14,7 +14,9 @@ dependencies {
     compile 'org.slf4j:slf4j-api:1.7.25'
     compile 'org.slf4j:slf4j-simple:1.7.25'
     compile 'com.typesafe:config:1.3.1'
-
+    
+    //jaxb is removed from JDK9+
+    //we need below jar in case to run on JDK8+
     compile 'javax.xml.bind:jaxb-api:2.2.12' 
 }
 
--- a/qcommon/src/org/qcmg/common/util/FileUtils.java
+++ b/qcommon/src/org/qcmg/common/util/FileUtils.java
@@ -64,7 +64,6 @@ public class FileUtils {
 	        ByteBuffer buff = ByteBuffer.allocate(2048);
 	
 	        while(channel.read(buff) != -1){
-	           // buff.flip();
 	        	//jdk8 don't have java.nio.ByteBuffer.flip()
 	        	//we have to cast it to Buffer, so it can run JDK8 and above
 	            ((Buffer)buff).flip();
--- a/qcommon/src/org/qcmg/common/util/FileUtils.java
+++ b/qcommon/src/org/qcmg/common/util/FileUtils.java
@@ -15,6 +15,7 @@ import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.nio.Buffer;
 import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
 import java.nio.file.FileVisitOption;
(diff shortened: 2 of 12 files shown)


qcommon/build.gradle
```diff
@@ -14,5 +14,7 @@ dependencies {
     compile 'org.slf4j:slf4j-api:1.7.25'
     compile 'org.slf4j:slf4j-simple:1.7.25'
     compile 'com.typesafe:config:1.3.1'
+
+    compile 'javax.xml.bind:jaxb-api:2.2.12'
```
holmeso: Is this used? Can't see any new imports in the changes that are coming from here.

qcommon/src/org/qcmg/common/util/FileUtils.java
```diff
@@ -61,8 +62,11 @@ public static String getFileCheckSum( final String file ) {
 	        MessageDigest md = MessageDigest.getInstance("MD5");
 	        FileChannel channel = inputStream.getChannel();
 	        ByteBuffer buff = ByteBuffer.allocate(2048);
+	
 	        while(channel.read(buff) != -1){
-	            buff.flip();
+	        	//jdk8 don't have java.nio.ByteBuffer.flip()
+	        	//we have to cast it to Buffer, so it can run JDK8 and above
+	            ((Buffer)buff).flip();
```
holmeso: you don't need to cast here - `ByteBuffer` is a subclass of `Buffer` and so can see its public and protected methods.
(just like your call to `clear()`)

ChristinaXu2017: Just made our tool more flexible, it can run on both JDK8 and above

qcommon/build.gradle
ChristinaXu2017: it is java version issue. just make our tool more flexible.

holmeso: I don't understand?

qcommon/src/org/qcmg/common/util/FileUtils.java
holmeso: Leaving the cast out should work fine in java8

qcommon/build.gradle
ChristinaXu2017: From JDK9 there is no JAXB.  Now many users use JDK9+, it is pain to ask users to swipe between java versions. To make our tools can run on java8+, it is better to add them during the build.

holmeso: I think changes for other java versions are out of scope for this PR and should be addressed in their own PR. And in this particular instance, I would presume that we would want to add this dependency at the top level rather than at the individual project level.

ChristinaXu2017: there is just one line change, it is not necessary to create another PR. It didn't affect anything, just make it more flexible run on JDK8 and above. We may move it to top-level but it will be next PR.

qcommon/src/org/qcmg/common/util/FileUtils.java
ChristinaXu2017: To cast  ByteBuffer to Buffer before calling flip(), Then users can compile the project with JDK9+ and have it running on JDK8.

qcommon/build.gradle
holmeso: I still believe that this is out of scope for this PR. It will result in users who are running on java8 needlessly having to download a redundant jar file that will not be used.
However, there is obviously some resistance to removing this dependency for whatever reason.
At the very least there would need to be some comments in the build file as to why this dependency is being added before I can approve

holmeso: Your comment is incorrect - you don't need the jaxb jar for jdk8

PR #150, opened 2020-01-15
name attribute in Kmers section should have different value for unPaired BAM and fastq Seq

ChristinaXu2017:
# Description

The value of name attribute is the length + "kmers" if there are unpaired reads. The reason was Fastq reads are all unPaired, but it won't make sense to call them  "unPaired reads".  However, we clarify reads to "firstReadInPair", "secondReadInPair" and "unPaired" in BAM files. It makes the output inconsistent, the kmers section output "kmers" but missing "unPaired" reads. 
Here we put an additional argument to KmersSummary.toXml(Element parent,  int klength , boolean isFastq). it will set "kmers" to attribute "name"  if the "isFastq" is true;  otherwise it will set "unPaired" for unPaired reads. 

## Type of change
- [ x] Bug fix (non-breaking change which fixes an issue)
- [x ] New feature (non-breaking change which adds functionality)
 
# How Has This Been Tested?
The unit test is updated, and extra code are also added. It also tested on BAM and Fastq file. 

# Checklist:

- [ X] My code follows the style guidelines of this project
- [ X] I have performed a self-review of my own code
- [ X] My changes generate no new warnings
- [ X ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/qprofiler2/src/org/qcmg/qprofiler2/bam/BamSummaryReport2.java
+++ b/qprofiler2/src/org/qcmg/qprofiler2/bam/BamSummaryReport2.java
@@ -196,8 +196,10 @@ public class BamSummaryReport2 extends SummaryReport {
 		}
 				
 		//1mers is same to baseByCycle
-		for( int i : new int[] { 2, 3, KmersSummary.maxKmers } )
+		for( int i : new int[] { 2, 3, KmersSummary.maxKmers } ) {
 			kmersSummary.toXml( parent,i, false);
+		}
+			
 	}
 	
 	//<QUAL>
--- a/qprofiler2/src/org/qcmg/qprofiler2/summarise/KmersSummary.java
+++ b/qprofiler2/src/org/qcmg/qprofiler2/summarise/KmersSummary.java
@@ -221,7 +221,6 @@ public class KmersSummary {
 			String name = BamSummaryReport2.sourceName[pair];
 			
 			//read may have no pair information such as fastq
-			//if(pair == 0 && parsedCount[1].get() == 0 && parsedCount[2].get() == 0)
 			if( isFastq )
 				name = klength+"mers";
 			Set<String> kmerStrs = getPopularKmerString(maxNo,  klength, false, pair);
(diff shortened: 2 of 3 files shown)


qprofiler2/src/org/qcmg/qprofiler2/summarise/KmersSummary.java
```diff
@@ -215,7 +221,7 @@ public void toXml( Element parent, int klength ) {
 			String name = BamSummaryReport2.sourceName[pair];
 			
 			//read may have no pair information such as fastq
-			if(pair == 0 && parsedCount[1].get() == 0 &&parsedCount[2].get() == 0)
+			if( isFastq )
```
holmeso: braces pls

qprofiler2/src/org/qcmg/qprofiler2/summarise/KmersSummary.java
```diff
@@ -126,8 +126,8 @@ public void parseKmers( byte[] readString, boolean reverse , int flagFirstOfPair
 		 for(int i = 0; i <= readString.length - merLength; i ++ ){
 			 byte[] mers = new byte[ merLength ];
 			 for(int j = 0; j <  merLength; j ++ )  mers[j] = dataString[ i + j ];	
-			 int pos = getPosition(i, mers );			 
-			 tally[flagFirstOfPair].increment(pos);
+			 int pos = getPosition( i, mers );
```
holmeso: braces in the line above pls

qprofiler2/src/org/qcmg/qprofiler2/bam/BamSummaryReport2.java
```diff
@@ -197,7 +197,7 @@ private void createSeq(Element parent){
 				
 		//1mers is same to baseByCycle
 		for( int i : new int[] { 2, 3, KmersSummary.maxKmers } )
-			kmersSummary.toXml( parent,i );
+			kmersSummary.toXml( parent,i, false);
```
holmeso: braces pls

PR #152, opened 2020-01-29
use MIN values to decide on MIUN annotation

### LINKED ISSUE #145: change default values for MIUN cutoff
At present, if 2 control reads that failed the filters contain the alt, then the `MIUN` annotation is added to the control filter column in the vcf record.

This is stricter that the case for reads which passed the filters, which require `max(2, 3%)` of control reads to have the alt before the `MIN` annotation is added.

It seems curious that failed filter reads are handled in a stricter manor than reads that have passed the filters.

I would suggest that the criteria for the application of the `MIUN` annotation should be at least as strict as for the `MIN` annotation if not more so.

And so I am proposing that the default value for the application of the `MIUN` annotation be changed to `max(3,5%)`.

An alternative would be to ignore the MIUN annotation when it comes to classifying a variant

This issue has been around for a while and recently appeared again when a control sample with coverage over 800 had the `MIUN` annotation because of 2 failed filter reads.

holmeso:
# Description

The `MIUN` (**M**utation **I**n **U**nfiltered **N**ormal) annotation is added when there is evidence of the alt in reads that failed the filters. This annotation is applied if the `MIN` annotation is not present.
Currently, there is a single hard cutoff for this annotation (defaults to 2). 
It turns out that this is stricter that the cutoff used by the `MIN `annotation (`max(2, 3% of reads)`), which uses better quality reads (ie. reads that have passed the filter).

This change addresses this discrepancy, and applies the same cutoffs that are used in the `MIN` annotation for the `MIUN` annotation.

The defaults are now:
```
hard cutoff = 2
percentage = 3%
```

This means that the failed filter reads will need to contain `max(hard filter, percentage) `copies of the alt before the `MIUN` annotation is added.

eg.
If the passed filter coverage is 90, and the failed filter field is "A2;C8" giving us a combined coverage of 100, and if the default values are used (max(2, 3%)) then if the alt field contained "C", we would get the `MIUN` annotation.
If the alt field did not contain "C" but rather contained "A", then we would not see the `MIUN` annotation (2 is less than max(2,3%)). 

The hard cutoff and percentage values are configurable by the user.

Fixes #145 

# How Has This Been Tested?

Additional unit tests have been provided to cover the updated and new code

# Checklist:

### DIFF
--- a/qannotate/src/au/edu/qimr/qannotate/modes/ConfidenceMode.java
+++ b/qannotate/src/au/edu/qimr/qannotate/modes/ConfidenceMode.java
@@ -55,6 +55,7 @@ public class ConfidenceMode extends AbstractMode{
 	public static final int sBiasAltPercentage = 5;
 	public static final int sBiasCovPercentage = 5;
 	
+	@Deprecated	// using values (both hard cutoff and percentage) from MIN for MIUN annotation
 	public static final int MIUN_CUTOFF = 2;	// based on existing values
 	
 	//filters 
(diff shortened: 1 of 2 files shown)


holmeso:
Thanks for the suggestion - easier for biologists to read!

PR #172, opened 2020-06-10
Qmotif create writer thread that runs in background

holmeso:
# Description

During a recent piece of work where `qmotif` was run against a thousand bam files, it became apparent that it was running out of memory.
These runs were different to normal in that we were looking at all regions of the genome rather than just the starts and ends of chromosomes as is usually done.

What was happening was that the SAMRecords that contained the motifs of interest were being stored in a queue until all the processing was done, and then written to file. With much larger regions than normal, this meant that a large number of SAMRecords were stored in memory, resulting in the `OutOfMemoryErrors` and high thread use due to GC.

The solution was to create a thread that writes the SAMRecords to file at the start of the process. This writer thread picks up the SAMRecords as they are placed into the queue meaning that they don't sit in memory.

A couple of other minor fixes have also been included in this PR:

- cached the hashCode value in `ChrPositionName` - this was using a (relatively) large amount of cpu time as `ChrPositionName` is a key in a widely used map. Caching this reduces cpu use here.
 - Changed from having a `StringBuilder` to represent the motifs found in a region to a map. This is really weird as judging by the code, maps were used originally and then changed to StringBuilder. This is really not sensible as when running with large regions, there will be a huge number of (duplicated) entries in the `StringBuilder`. Reverted back.
 - fixed some styling on classes that were updated

## Type of change

# How Has This Been Tested?
Have run the updated code and compared against an existing output, and results were the same.

# Checklist:

### DIFF
--- a/qcommon/src/org/qcmg/common/model/ChrPositionName.java
+++ b/qcommon/src/org/qcmg/common/model/ChrPositionName.java
@@ -23,6 +23,7 @@ public class ChrPositionName extends ChrRangePosition implements ChrPosition {
 	
 	
 	private final String name;
+	private volatile int hashCode;
 
 	public ChrPositionName(String chromosome, int position, int endPosition, String name) {
 		super(chromosome, position, endPosition);
@@ -53,9 +54,14 @@ public class ChrPositionName extends ChrRangePosition implements ChrPosition {
 	 */
 	@Override
 	public int hashCode() {
-		final int prime = 31;
-		int result = super.hashCode();
-		result = prime * result + ((name == null) ? 0 : name.hashCode());
+	    int result = hashCode;
+	    if (result == 0) {
+	    	result = 17;
+			final int prime = 31;
+			result = super.hashCode();
+			result = prime * result + ((name == null) ? 0 : name.hashCode());
+			hashCode = result;
+	    }
 		return result;
 	}
 	/* (non-Javadoc)
--- a/qcommon/src/org/qcmg/common/model/ChrPositionName.java
+++ b/qcommon/src/org/qcmg/common/model/ChrPositionName.java
@@ -56,7 +56,7 @@ public class ChrPositionName extends ChrRangePosition implements ChrPosition {
 	public int hashCode() {
 	    int result = hashCode;
 	    if (result == 0) {
-	    	result = 17;
+			result = 17;
 			final int prime = 31;
 			result = super.hashCode();
 			result = prime * result + ((name == null) ? 0 : name.hashCode());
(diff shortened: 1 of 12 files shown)


qcommon/src/org/qcmg/common/model/ChrPositionName.java
```diff
@@ -53,9 +54,14 @@ public String toIGVString() {
 	 */
 	@Override
 	public int hashCode() {
-		final int prime = 31;
-		int result = super.hashCode();
-		result = prime * result + ((name == null) ? 0 : name.hashCode());
+	    int result = hashCode;
+	    if (result == 0) {
+	    	result = 17;
+			final int prime = 31;
```
ChristinaXu2017, line 60: why bigger indent than last line?

qmotif/src/org/qcmg/motif/JobQueue.java
```diff
@@ -146,35 +148,34 @@ public JobQueue(final Configuration invariants)
...
-		writeOutput();
-		logger.info("DONE - Writing output");
-		
 		logger.info("total number from input is " + countIn.get());
 		logger.info("total number of reads (satisfied by the query if query provided) count to coverage is " + countOut.get());
 		
 		logger.info("Performing final reduce step on results");
 		reduceResults();
 		logger.info("Final reduce step complete");
 //		logger.debug("Final reduced results: " + perIdPerCoverageBaseCounts);
```
ChristinaXu2017, line 173: delete it

qmotif/test/org/qcmg/motif/util/RegionCounterTest.java
```diff
@@ -45,17 +45,17 @@ public void excludedRegion() {
 	public void includedRegionUnmappedReadsFS() {
 		rcInc.addMotif("ABCD", true, false);
 		assertEquals(true, rcInc.hasMotifs());
-//		assertEquals(true, rcInc.getMotifsForwardStrand().containsKey("ABCD"));
-		assertEquals("ABCD", rcInc.getMotifsForwardStrand());
+		assertEquals(true, rcInc.getMotifsForwardStrand().containsKey("ABCD"));
+//		assertEquals("ABCD", rcInc.getMotifsForwardStrand());
```
ChristinaXu2017, line 49: should this commented line be removed?

holmeso: done

qmotif/src/org/qcmg/motif/JobQueue.java
holmeso: done

qcommon/src/org/qcmg/common/model/ChrPositionName.java
holmeso: fixed

holmeso:
thanks for your review Christina

Issue #178, opened 2020-06-25
TemporaryFolder issue reported by spotbugs


ChristinaXu2017:
**Describe the bug**
TemporaryFolder rule was used in our unit test. it allows the creation of files and folders that should be deleted when the test method finishes (whether it passes or fails). Whether the deletion is successful or not is not checked by this rule. No exception will be thrown in case the deletion fails. However, the TemporaryFolder rule must not be static but the many methods in the Junit test is static. 

**To Reproduce**
Before we have to work around, such as, create a temporary in the "@Before" method. Lucky after  junit4.9,  there is " ClassRule", it allows the TemporaryFolder instance to be static. 
eg.  In the below example, we can only create the input file once, and then it can be used by all test methods

 ```
@ClassRule	
 public  static TemporaryFolder testFolder = new TemporaryFolder();
 private File input;

@BeforeClass
 public static void createInput throws Exception { 
		input = testFolder.newFile("testInputFile.sam");
		//...
}
```

Issue #177, opened 2020-06-25
encoding issue


ChristinaXu2017:
**Describe the bug**
After running the tool called "spotbugs", we get below warning message:

-  Format string should use %n rather than \n

In format strings, it is generally preferable to use %n, which will produce the platform-specific line separator.
%n is portable between various platforms, the value emitted from %n will suit the underlying platform, whereas value emitted by \n is same for all the platforms.\n is the correct newline character for Unix-based systems, other systems may use different characters to represent the end of a line. Windows system use \r\n, and early MacOS systems used \r.

-  Reliance on default encoding

eg. org.qcmg.qprofiler2.cohort.CohortSummarizer.outputSumamry, it was
`try (BufferedWriter writer =  new BufferedWriter(new FileWriter(output))) {  … }`
Here FileWriter is too convenient and spotBugs is not happy about it. Further, there is no way to set a charset using FileWriter. we can use OutputStreamWriter instead because it allows us to specify a charset. UTF-8 is a good choice because it is always supported and can encode any character.

eg. `byte[] readBases = record.getReadString().getBytes(Charset.forName("UTF-8"));`

**To Reproduce**
 The above issues are resolved in the coming pull request.

Issue #176, opened 2020-06-25
 Inefficient  code reported by spotbugs


ChristinaXu2017:
**Describe the bug**
After running the tool called "spotbugs", we get below warning message:

- Inefficient use of keySet iterator instead of entrySet iterator
eg. In method org.qcmg.qprofiler2.bam.BamSummaryReport2.createRNAME(Element)
`for ( String chr : rNamePosition.keySet()) { PositionSummary posSum = rNamePosition.get(chr);… } `
should change to 
`for ( Entry<String, PositionSummary> entry : rNamePosition.entrySet() ) { PositionSummary posSum = entry.getValue();… }`

-  Should be a static inner class
eg. org.qcmg.qprofiler2.cohort.CohortSummaryReport$Category is an inner class but it didn’t access the outer, as basic rule, we should make it static by default.

- Write to static field from instance method
eg. In org.qcmg.qprofiler2.QProfiler2
`private static String[] cmdLineFiles; public int setup(String args[]) { cmdLineFiles = options.getFileNames(); …}`
Here the class method is access the static field, one options is to remove static for cmdLineFiles if it is not accessed by outer. 

-  Switch statement found where default case is missing
eg. org.qcmg.qprofiler2.summarise.KmersSummary.getEntry, The default case was missing. 

-  Method invokes System.exit(...)
eg. new org.qcmg.qprofiler2.summarise.KmersSummary(int) {… System.exit(1);}
Invoking System.exit shuts down the entire Java virtual machine. Consider throwing a RuntimeException or using "return" in a constructor. 

-  Method ignores return value
eg. In class org.qcmg.qprofiler2.fastq.FastqSummarizerMT,  the return value from java.util.concurrent.CountDownLatch.await(timeout) is ignored. However there is no need to receive the value, so we have to ignore this bug report. 

**To Reproduce**
The above issues are resolved in the coming pull request.

Issue #175, opened 2020-06-25
array issues reported by spotbugs


ChristinaXu2017:
**Describe the bug**
After running the tool called "spotbugs", we get below warning message:

- May expose internal representation by incorporating reference to mutable object
eg. `new org.qcmg.qprofiler2.vcf.VcfSummarizer(String[] formatArgs){ this.formats = formatArgs }.` 
Java array is immutable,  it pass a reference to the method and the reference is untrusted. Storing a copy of the object is better approach, so change it to 
`{this.formats = Arrays.copyOf( formatArgs, formatArgs.length );}
`
-  Field is a mutable array
eg. org.qcmg.qprofiler2.bam.BamSummaryReport2.sourceName is a mutable array but set to final. Here we should use unmodifiableList
`public static final String[] sourceName = new String[]{
 "unPaired", XmlUtils.FIRST_PAIR, XmlUtils.SECOND_PAIR
 };`
It should be
`public static final List<String> sourceName = Collections.unmodifiableList
(Arrays.asList("unPaired", XmlUtils.FIRST_PAIR, XmlUtils.SECOND_PAIR));`

-  May expose internal representation by returning reference to mutable object
eg. org.qcmg.qprofiler2.summarise.KmersSummary.getPossibleKmerString(…)
`{ String[] mers = str1.split(Constants.COMMA_STRING);… return mers}`
Here return a reference of mutable array, and he reference is untrusted. Storing a copy of the object is better approach. It should change to 
`{…return Arrays.copyOf( mers, mers.length );}`

- The field should be package protected
eg. In org.qcmg.qprofiler2.summarise.KmersSummary, the array is mutable but set to final, so get a warning “A mutable static field could be changed by malicious code or by accident.”
`public static final char[] atgcCharArray = new char[]{'A','T','G','C'}; `change to 
`public static final List<Character>  atgcCharArray = Collections.unmodifiableList(Arrays.asList('A','T','G','C'));`

**To Reproduce**
The above issues are resolved in the coming pull request.

PR #182, opened 2020-07-22
Bug spot

### LINKED ISSUE #174: variable issue reported by spotbugs
**Describe the bug**
After running the tool called spotbugs, we get below warning message:

- Dead store to the local variable
eg. Variable “pgLines” in org.qcmg.qprofiler2.bam.BamSummarizer2::createReport is stored value but not used,  it should be deleted. 

-  Unread public/protected field
eg. org.qcmg.qprofiler2.summarise.PairSummary.logger is not used and should be delete.

- Unread field
eg. field: org.qcmg.qprofiler2.summarise.SampleSummary.sampleTrans, is not read and should be deleted.

- Field isn't final but should be
eg. org.qcmg.qprofiler2.cohort.CohortSummaryReport.headerline isn't final but should be.

- Boxed value is unboxed and then immediately reboxed
eg org.qcmg.qprofiler2.util.XmlUtils.outputValueNode(Element parent, String name, Number value)
`{ String v = String.valueOf(value); … v = String.format("%,.2f", (double)value); } ` 
should use auto box and change to
`{ String v = value + “”; … v = String.format("%,.2f", value);`

-  Method uses the same code for two branches
eg. new org.qcmg.qprofiler2.summarise.CycleSummary(Object, int) uses the same code for two branches, the code was
```
int DEFAULT_NO_OF_KEYS_CHARACTER = 128;
int DEFAULT_NO_OF_KEYS_INTEGER = 128;
int noOfKeys = type instanceof Character? DEFAULT_NO_OF_KEYS_CHARACTER : DEFAULT_NO_OF_KEYS_INTEGER;
```
 It should change to one line
`int noOfKeys = 128;`

**To Reproduce**
Above issues are resolved in the comming pull request.

ChristinaXu2017:
# Description

run spotbugs on qcommon, most related bugs are fixed.  Fixes #174 #175 #176 #177 #178. However some problems are ignored,  such as missing default cases in switch statement, call start() method in threads. etc
 
## Type of change

- [X ] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

used existing unit test code

# Checklist:

- [X ] My code follows the style guidelines of this project
- [X ] I have performed a self-review of my own code
- [X ] My changes generate no new warnings
- [X ] I have added tests that prove my fix is effective or that my feature works


holmeso:
Could you pls use StringBuilder rather than StringBuffer - its quicker (doesn't synchronise)

holmeso:
Why the move from static methods to instance methods in FisherExact class?

qcommon/src/org/qcmg/common/model/PileupElement.java
```diff
@@ -73,12 +76,30 @@ public int getTotalCount() {
...
 	}
-
+	@Override
 	public int compareTo(PileupElement o) {
 		// only interested in the total count for the purposes or ordering
 		return o.getTotalCount() - getTotalCount();
 	}
 	
+	//you need to implement the equals method thus overriding the default compareTo
+	@Override
```
holmeso, line 86: this equals and hashcode methods look a little sparse.
Eclipse provides a decent implementation of these methods which will be a little more robust.
Right click in the class, go to `Source` -> `Generate hashCode and equals()`

qcommon/src/org/qcmg/common/model/PileupElementLite.java
```diff
@@ -118,5 +118,24 @@ public int compareTo(PileupElementLite o) {
 		// only interested in the total count for the purposes or ordering
 		return o.getTotalCount() - getTotalCount();
 	}
+	
+	//you need to implement the equals method thus overriding the default compareTo
+	@Override
```
holmeso, line 123: need a stronger equals and hashCode

qcommon/src/org/qcmg/common/util/AccumulatorUtils.java
```diff
@@ -464,6 +465,33 @@ else if (isRef)
 				return -1;
 			return o.quality - quality;
 		}
+		
+		//you need to implement the equals method thus overriding the default compareTo
+		@Override 
+		public boolean equals(Object o) {
```
holmeso, line 471: don't think we need equals and hashcode for a utility class with only static methods

qcommon/src/org/qcmg/common/util/LoadReferencedClasses.java
```diff
@@ -20,8 +20,8 @@
 	
 	
 	public static final <T> void loadClasses(final Class<T> clazz) throws URISyntaxException, IOException, ClassNotFoundException {
-		final long start = System.currentTimeMillis();
-		final long mem = Runtime.getRuntime().freeMemory();
+//		final long start = System.currentTimeMillis();
```
holmeso, line 23: delete

qcommon/src/org/qcmg/common/model/PileupElement.java
ChristinaXu2017: thanks for suggestion

qcommon/src/org/qcmg/common/model/PileupElementLite.java
ChristinaXu2017: replaced by eclipse provided one

qcommon/src/org/qcmg/common/util/AccumulatorUtils.java
ChristinaXu2017: I agree, and removed

qcommon/src/org/qcmg/common/util/LoadReferencedClasses.java
ChristinaXu2017: tidied up

holmeso:
FisherExact should really be deprecated and then removed in a future release

PR #183, opened 2020-08-05
Cram type

ChristinaXu2017:
# Description

Here we allow qprofielr2 to take cram file as input. The process will be the same as the BAM/SAM mode, but the required reference file have to set to the environment variable, eg.  -Dsamjdk.reference_fasta=/reference/genomes/GRCh37_ICGC_standard_v2/indexes/BWAKIT_0.7.12/GRCh37_ICGC_standard_v2.fa.

We also created a CramFileReaderFactory, but there are still many investigations required before qprofiler2 can utilize this class directly.  We are also not sure how many more cram file come to our center, that means we are not sure whether it is worth to spend more time to continue this cram reader. 

## Type of change

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [X] Test A: updated current unit test 
- [X] Test B: run on a real cram file. 

# Checklist:


qpicard/src/org/qcmg/picard/CramFileReaderFactory.java
```diff
@@ -0,0 +1,103 @@
+/**
```
holmeso, line 1: don't need this file - can read CRAM files using existing classes with the relevant java option set. Also, if we did want to go down the route of not having to set the java option, we should be extending the existing classes

qcommon/src/org/qcmg/common/model/ProfileType.java
```diff
@@ -101,7 +103,7 @@ public static ProfileType getType2(String s) {
 	    	ext = s.substring(i+1).toLowerCase();
 	    }
 	    
-	    if (BAM_EXTENSION.equals(ext) || SAM_EXTENSION.equals(ext))
+	    if (BAM_EXTENSION.equals(ext) || SAM_EXTENSION.equals(ext) || ext.contentEquals(CRAM_EXTENSION))
```
holmeso, line 106: why contentEquals and not equals like the others?

ChristinaXu2017: typo fixed

qpicard/src/org/qcmg/picard/CramFileReaderFactory.java
ChristinaXu2017: deleted

qcommon/src/org/qcmg/common/model/ProfileType.java
```diff
@@ -101,7 +103,7 @@ public static ProfileType getType2(String s) {
 	    	ext = s.substring(i+1).toLowerCase();
 	    }
 	    
-	    if (BAM_EXTENSION.equals(ext) || SAM_EXTENSION.equals(ext))
+	    if (BAM_EXTENSION.equals(ext) || SAM_EXTENSION.equals(ext) || ext.equals(CRAM_EXTENSION))
```
holmeso, line 106: if ext is null here, then a NPE will be thrown, whereas if you do `CRAM_EXTENSION.equals(ext)` then it will never throw a NPE because `CRAM_EXTENSION` is a non-null constant

ChristinaXu2017: I agree and fixed

PR #189, opened 2020-09-15
qsv - renamed PairedEndRecord to ZPAnnotator

### LINKED ISSUE #188: qsv - change PairedEndRecord to a utility class

If running in "pe" mode, QSV currently creates a `PairedEndRecord` for each paired `SAMRecord`.
This object is created, a tally of the ZP annotation is kept, the `SAMRecord` has an attribute set, and then the object is set to null.

```
PairedEndRecord pe = new PairedEndRecord(record, lower, upper);
pe.createZPAnnotation();					
countZp(pe.getZPAnnotation());
pe = null;
```

Each paired record in a BAM is currently put through this, which for large BAMs amounts to a fair amount of work for the garbage collector.

I propose that rather than have a `PairedEndRecord` class, we instead have a `ZPAnnotator` class that takes in a SAMRecord along with the lower and upper bounds (both ints), and returns the String annotation.
This can then be used to set the ZP attribute in the SAMRecord

Do nowt

holmeso:
# Description

`PairedEndRecords` were being instantiated just to extract a String that was being used to set an attribute (ZP) within a SAMRecord.
The object is not persisted and is not required. The only thing of interest is the String annotation.
This PR removes the need to instantiate the `PariedEndRecord`, instead providing a static method that will return a String.
The `PairedEndRecord` class has also been renamed to `ZPAnnotator` to more accurately reflect the purpose of the class.

Fixes #188 

# How Has This Been Tested?

I have build a jar using these changes and have run against some test BAMs. Outputs are identical to output produced by current code.

Existing unit tests have been updated to reflect the code changes and pass.

# Checklist:


ChristinaXu2017:
The code looks good, but you should edit this pull request description.

holmeso:
> The code looks good, but you should tidy up this pull request description.

How so?

ChristinaXu2017:
"Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change." should be removed or not?

holmeso:
> "Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change." should be removed or not?

You're right - deleted

ChristinaXu2017 (approved):

PR #190, opened 2020-09-15
qsv - reduce the number of calls to SAMRecod.getReadGroup()

holmeso:
# Description

When profiling `qsv`, it became apparent that there were a large number of calls to `SAMRecord`'s `getReadGroup` method.
This method is non trivial as it needs to find the `RG` attribute before then calling a lookup on a map.

This PR attempts to reduce the calls to this method, instead caching the value locally, and using this value where required.

eg.
Instead of :
```
String key = "blah" + record.getReadGroup() != null ? record.getReadGroup().getId() :  "";
```
we have:
```
SAMRecordReadGroup srrg = record.getReadGroup();
String key = "blah" + srrg != null ? srrg.getId() : "";
```

# How Has This Been Tested?

All existing unit test pass.
Changes have been tested against test BAM files and output is identical to current code

# Checklist:


ChristinaXu2017:
I checked the code inside SAMRecord.class, it workaround to get the readGroupId, because it is not a compulsory field in SAMRecord.  How about we just check the RG tag from each read record but not call the method SAMRecord.getReadGroup().getId().  It should save a huge amount of computational time to get the readGroupId string. that is changing
          SAMRecord.getReadGroup().getId()
to
          SAMRecord. getAttribute(SAMTag.RG.name())

ChristinaXu2017:
alternatively, you can create a wrapper, it contains the SAMRecord and readGroupId. It creates garbage collection of objects during SAMRecordIterator, however, it is trivial compare to the seeking readGroupId by the method of SAMRecord.getReadGroup().getId().

holmeso:
> I checked the code inside SAMRecord.class, it workaround to get the readGroupId, because it is not a compulsory field in SAMRecord. How about we just check the RG tag from each read record but not call the method SAMRecord.getReadGroup().getId(). It should save a huge amount of computational time to get the readGroupId string. that is changing
> SAMRecord.getReadGroup().getId()
> to
> SAMRecord. getAttribute(SAMTag.RG.name())

Good suggestion, but I would rather not use a different value of RG ID to what is currently being used.

holmeso:
> alternatively, you can create a wrapper, it contains the SAMRecord and readGroupId. It creates garbage collection of objects during SAMRecordIterator, however, it is trivial compare to the seeking readGroupId by the method of SAMRecord.getReadGroup().getId().

I'm not so keen on this solution. There may be instances where the read group id is not required, and you are then forcing the caching of this value. Also, what would happen if there were other methods (eg. getReadString) that are computationally intensive that you would like to capture. Does that mean you would need another wrapper class?

Most of the objects that are being created with a SAMRecord and read group id are using a small subset of the data in the SAMRecord, and it would make more sense (from a GC perspective) to split them out and pass individually.

ChristinaXu2017:
the method SoftClipStaticMethods :: createSoftClipRecord(SAMRecord record, String rgId, Integer start, Integer end, String chromosome)  should change to

holmeso:
> the method SoftClipStaticMethods :: createSoftClipRecord(SAMRecord record, String rgId, Integer start, Integer end, String chromosome) should change to
> createSoftClipRecord(SAMRecord record, String rgId, int start, int end, String chromosome)
> 
> the method QSVUtils::createRecord(int bpPos, Integer chrStart, Integer chrEnd) should also change to
> createRecord(int bpPos, int chrStart, int chrEnd)
> 
> If you check all call hierarchy, these methods repeatedly autoboxing, unboxing between Integer and int. Actually, there should have no Integer involved.
> 
> AnotateFilterMT::pileupSoftClips(...int start, int end ...) calls above methods directly and indirectly.

changes made as requested

ChristinaXu2017 (approved):

PR #191, opened 2020-09-15
order chromosomes for annotation

holmeso:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Qsv's `AnnotateFilterMT` class reads the records in a BAM file. It creates multiple threads that take a chromosome from a queue and reads the records from the BAM file for that chromosome.
The order in which the chromosomes are added to the queue is not currently ordered, meaning that it is possible for `chr1` (typically the chromosome with the most reads) to be handled late in the process, which means that the overall time taken to process the BAM file could be larger than it needs to be.

This PR sorts the list of chromosomes in the BAM file before adding them to the queue. This ensures that the larger chromosomes are handled first, which should reduce the time spent reading from the BAM files.

# How Has This Been Tested?

Existing unit tests pass.
When testing against smaller BAMs, it was confirmed that the chromosomes were processed in the required order.
Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any 

# Checklist:

### DIFF
--- a/qsv/src/org/qcmg/qsv/annotate/AnnotateFilterMT.java
+++ b/qsv/src/org/qcmg/qsv/annotate/AnnotateFilterMT.java
@@ -20,17 +20,9 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


ChristinaXu2017 (approved):
looks good

PR #194, opened 2020-09-30
Q3tiledaligner refactor

holmeso:
# Description

This PR contains the bulk of the `qsv` tiled aligner work. The code added in this PR will be used (potentially) by `qsv` to align sequences to the genome.
Note that this PR does not change `qsv` from using `BLAT` to using the `tiled aligner`, rather, it puts the code in place for a future transition once sign-off has been obtained.

Key features:
 - `TiledAligner` class allows the user to attempt to align a sequence of interest using the tiled aligner
 - `GenerateTiledAlignerFile` allows user to create a tiled aligner file based on a supplied reference file
 - Programatic access to tiled aligner provided by `TiledAlignerUtil` class 

## Type of change

# How Has This Been Tested?

A large number of new unit tests have been added.
Existing unit tests pass

# Checklist:


ChristinaXu2017 (commented):
you added a new package org.qcmg.string into qio, it seems they are not yet called by anyone.  They are similar to normal inputStream and outputStream but can read file header. The file header is also ambiguous, it suppose to be a string starting with "#", but the "#" is not compulsory while adding the header into a file.

holmeso:
> you added a new package org.qcmg.string into qio, it seems they are not yet called by anyone. They are similar to normal inputStream and outputStream but can read file header. The file header is also ambiguous, it suppose to be a string starting with "#", but the "#" is not compulsory while adding the header into a file.

I've updated the package so that the StringFileReader constructor can be passed a CharSequence that defines the start of a header line.
This makes the package more generic and open to use by anyone.
I'm not sure it's the job of the Writer to enforce any kind of format upon the header - that should surely be up to the user.

ChristinaXu2017 (approved):
Let's merge to master first, Then we should think about re-organize the qio, there are too many similar packages.

Issue #198, opened 2020-10-22
encoding issue in qio


ChristinaXu2017:
**Is your feature request related to a problem? Please describe.**
Every implementation of the Java platform is required to support the following standard charsets: US-ASCII, ISO-8859-1, UTF-8/16/16BE/16LE.  Every instance of the Java virtual machine has a default charset,  java one-argument constructors of IO class, such as, FileReader, always use the platform default encoding which is generally a bad idea.  

The qio is a collection of File reader/writer for our internal files, such as GFF, VCF, MAF etc. So far we used default encoding, but we should set it to UTF-8, due to these internal files including gzip files are pure English character, even the description in BAM file are UTF-8. 

**Describe the solution you'd like**
All constructors in qio use default UTF-8 if it is single argument method; we also allow user to set charset by multi-argument constructors.

Issue #197, opened 2020-10-22
buffered IO vs stream IO


ChristinaXu2017:
**Is your feature request related to a problem? Please describe.**
The qio is a collection of File reader/writer for our internal files, such as GFF, VCF, MAF etc.  Some use buffered IO but some use stream, the read() or flush() method in the stream is a heavy operation, buffered IO will be much faster to process our internal files. 

**Describe the solution you'd like**
create a generic buffered reader/writer to process our tab-delimited files, such as GFF, VCF, germlineDB, gff3, Illumina, dbSnp, pileup file, etc. 

**Describe alternatives you've considered**
this generic IO class is only for the text file including gzip file. Other format files, such as BAM, XML, should use other method.

PR #209, opened 2020-11-04
create universal reader and writer

### LINKED ISSUE #196: passing outer stream into closable FileReader constructor
In our current `public final class VCFFileReader implements Closeable, Iterable<VcfRecord>`, it takes outer created inputStream into the constructor `public VCFFileReader(final InputStream instrm, Integer headerMaxBytes) `.  Here the InputStream has to support marking, such as BufferedInputStream; this stream will auto close in a try-with-resource block.  At moment these two restricts are ambiguously required, it causes exceptions.  eg. 
`BufferedInputStream is = new BufferedInputStream(fis);`
`try(VCFFileReader reader = new VCFFileReader(is)) { } catch(Exception ex) {} ` 
`System.out.println(is.read()); //get "Stream Closed" error`

It may be common in many JDK classes that take stream or reader from outside the class. However, our AdamaJava always uses "try with resources statement" to create VCFFileReader instance.  Hence, we should think about creating the BufferedInputStream inside VCFFileReader.  It will be replaced by a new constructor `public VCFFileReader(final File file, int bufferSize)`, it creates an internal BufferedInputStream.

We may deprecate this constructor at moment, in case it is desired in some project. such as multi-threads processing a stream.

ChristinaXu2017:
## Description

Our current qio project contains more than 20 different file readers and writers, most of them are similar. Some of them are hardly maintained due to old or deprecated file format. Here a universal file reader and writer is created, which can cover most file format, and also can be used as a base class for special and the complicated file format, such as VCF, FASTA, GFF, Illumina, etc. 

Fixes  #196 #197 #198 

## Type of change

# How Has This Been Tested?

A unit test class is added.  At moment these three class is not yet used in any projects, it will be further tested after replace current old other reader and writer

# Checklist:


qio/src/org/qcmg/qio/record/RecordReader.java
```diff
@@ -0,0 +1,137 @@
...
+		}
+    }
+    /**
+     * this method is overridable in subclass, eg illumina file have different header patten
+     * 
+     * @param headerPrefix
+     * @return the first line just after header
+     * @throws IOException
+     */
+    public String readHeader(CharSequence headerPrefix ) throws IOException{
```
holmeso, line 70: maybe change to more meaningful name ie. `readHeaderAndReturnFirstNonHeaderLine`

qio/src/org/qcmg/qio/record/RecordReader.java
```diff
@@ -0,0 +1,137 @@
...
+            	
+            	try {  
+            		//get next record, it may read  multi lines
+            		String line = bin.readLine();           		
+            		if(line != null ) {
+            			next = getRecord( line );
+            		}
+            		
+           			return rec;
+        		} catch (Exception e) {
```
holmeso, line 124: should avoid catching Exception - be more specific

holmeso (changes requested):

qio/src/org/qcmg/qio/record/RecordReader.java
```diff
@@ -0,0 +1,145 @@
...
+    public String readHeaderAndReturnFirstNonHeaderLine(CharSequence headerPrefix ) throws IOException{
+    	String nextLine = bin.readLine();
+    	
+    	//empty file
+    	if( nextLine == null ) return null;   	
+   	
+    	if(headerPrefix == null) return nextLine;   	 
+    	
+		//reader header, hence file pointer to first line after header
+		while ( null != nextLine && nextLine.startsWith(headerPrefix+"") ) {
```
holmeso, line 82: null guard already been performed

qio/src/org/qcmg/qio/record/StringFileReader.java
```diff
@@ -0,0 +1,30 @@
...
+	}
+	
+	public StringFileReader(File file, int bufferSize) throws IOException {
+		super(file, bufferSize);
+	}
+
+	@Override
+	public String getRecord(String line) {
+		if( null == line) {
+			throw new NullPointerException("can't convert null string to record!");
```
holmeso, line 26: Could this be `IllegalArgumentException` instead of NPE?

holmeso (changes requested):

qio/src/org/qcmg/qio/record/RecordReader.java
```diff
@@ -0,0 +1,147 @@
...
+        InputStream inputStream =  (isGzip) ? new GZIPInputStream(new FileInputStream(file)) : new FileInputStream(file);  
+        InputStreamReader streamReader = new InputStreamReader(inputStream, charset);
+        bin = new BufferedReader(streamReader, bufferSize);
+               		
+        String nextLine = readHeaderAndReturnFirstNonHeaderLine(headerPrefix);//bin.readLine();        
+		
+		try {
+			//get first record, set to null for empty file
+			next = nextLine == null? null : getRecord(nextLine);
+		}catch(Exception e) {
```
holmeso, line 62: IS there a more specific exception that could be caught here? I am always a little uncomfortable with `catch (Exception e) ` - it should be used as a last resort IMO

qio/src/org/qcmg/qio/record/RecordReader.java
```diff
@@ -0,0 +1,137 @@
...
+		} 
+		
+		return nextLine;
+    }
+    
+ /**
+  * This reader can maxmum take Integer.max lines of file header. Please make other header if bigger than this. 
+  * @return a list of header lines
+  */
+    public List<String>  getHeader() { return headerLines; }
```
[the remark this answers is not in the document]
holmeso: does this pass checkstyle's checks?

holmeso (approved):
LGTM

PR #215, opened 2020-11-24
restructure qio project

ChristinaXu2017:
# Description

In this pull request, we restructure qio project in below ways:
1. move unused old classes from  qio::org.qcmg.bed.*, qio::org.qcmg.consensuscalls.*, qio::org.qcmg.maf.*, qio::org.qcmg.primer*.* , qio::org.qcmg.string.*  to qio::org.qcmg.unused.

2. move old classes used by qmule from qio::org.qcmg.chrconv.*,  qio::org.qcmg.dbsnp.*, qio::org.qcmg.germlinedb.*, qio::org.qcmg.pileup.VerifiedSnp* to qmule::org.qcmg.qmule.io. 

3. use the new qio::org.qcmg.qio.record.StringFileReader to replace the old qio::org.qcmg.string.StringFileReader inside q3TiledAligner. 

Here 59 files are modified just because of class relocation, but no other code is modified. 

## Type of change

# How Has This Been Tested?

We use the existing unit tests plus regression tests.  We run the new q3tiledAligner on chr1 and compare it to the previous q3tiledAligner from adamajava/nightly.  We have to make sure the output from both versions is identical but the new one should use less time under the same resource required. 

# Checklist:


holmeso:
why not just deprecate packages that are no longer used rather than moving them into a new `unused` package?

ChristinaXu2017:
> why not just deprecate packages that are no longer used rather than moving them into a new `unused` package?

There are too many files, it is really a mess if just marked as deprecated. see the example in qSignature, there are so many relationships between deprecated and un-deprecated classes, methods. It is very hard to maintain them.

ChristinaXu2017:
> > why not just deprecate packages that are no longer used rather than moving them into a new `unused` package?
> 
> There are too many files, it is really a mess if just marked as deprecated. see the example in qSignature, there are so many relationships between deprecated and un-deprecated classes, methods. It is very hard to maintain them.

It is possible to remove them back or mark them as deprecated in the next several branches after qio is upgraded.  At moment it is easy to classify the relationship between these classes.

holmeso:
> > why not just deprecate packages that are no longer used rather than moving them into a new `unused` package?
> 
> There are too many files, it is really a mess if just marked as deprecated. see the example in qSignature, there are so many relationships between deprecated and un-deprecated classes, methods. It is very hard to maintain them.

It seems strange to me to be creating a new package for code that is not being used. It would make more sense to create a new package for the new code, leave the old code in the existing packages, and deprecate/remove when they are no longer required.

ChristinaXu2017:
> > > why not just deprecate packages that are no longer used rather than moving them into a new `unused` package?
> > 
> > 
> > There are too many files, it is really a mess if just marked as deprecated. see the example in qSignature, there are so many relationships between deprecated and un-deprecated classes, methods. It is very hard to maintain them.
> 
> It seems strange to me to be creating a new package for code that is not being used. It would make more sense to create a new package for the new code, leave the old code in the existing packages, and deprecate/remove when they are no longer required.

I thought about it and prefer in this way but was worried you may not happy with it.

ChristinaXu2017:
this unused package is harmless, won't affect any project. How about let them stay there at moment. let's focus on update qio, meanwhile, think about a new home for these unused classes.

holmeso:
> this unused package is harmless, won't affect any project. How about let them stay there at moment. let's focus on update qio, meanwhile, think about a new home for these unused classes.
> 
> It is just part of qio upgrading, there are more classes that should be removed or replaced. finally, I can move them back to the original location and mark them as deprecated. After the next release, we can delete these packages and classes.
> 
> alternatively, I can stop this pull request. after I finished all updates on qio, let you review. These unused classes will be marked as deprecated. But at that time, there will have hundreds of changed classes since almost every project in our adamaJava will be updated with new qio. Only qmule use old qio classes.

Maybe this PR should just be concerned with `qmule` and its use of the `qio` package?

ChristinaXu2017:
it is replaced by Bdevelop.qio

PR #221, opened 2020-12-09
abort to output ACLAP columns on vcf2maf mode

### LINKED ISSUE #220: qannotate vcf2maf is always including ACLAP fields, but not supplying the headers
**Describe the bug**
All maf records contain the `Note_Test_ACLAP` and `Note_Control_ACLAP` fields, but the headers are only sometimes supplied.

**To Reproduce**
1. Go to '/working/genomeinfo/cromwell-test/cromwell-executions/somaticDnaFastqToMaf/8177bed6-c4ba-4da9-a3d5-d4fc8d7a3e7b/call-vcf2mafAnno/execution/aa4149de-e2c3-4792-8905-d70a0cda060b.Somatic.Pass.maf'
2. Look at the end of each record, and the Note_Test_ACLAP and Note_Control_ACLAP fields are there.

**Expected behavior**
Maf records should only contain the `Note_Test_ACLAP` and `Note_Control_ACLAP` fields when they are also in the header.
Alternatively, always supply the header details for these fields.

**Additional context**
It looks like there is some logic in `Vcf2Maf.java` that checks the vcf file for an attribute, and if present, adds the headers.
There is no such logic when printing the maf record to file (although the ability to ignore these fields is present)

```
//if vcf header contain ACSNP descripion  then we have to output to maf file
if ( header.getFormatRecord(VcfHeaderUtils.FORMAT_ACLAP) != null ) {
	hasACLAP = true;
}
 			
for (MafElement ele: EnumSet.allOf( MafElement.class)) {
	if (hasACLAP || (!hasACLAP && ele.getColumnNumber() < 63)) {
		write.println(ele.getDescriptionLine());
	}
}
```

ChristinaXu2017:
# Description

qannotate vcf2maf mode never check ACLAP format field on input vcf file but output null value to "Note_test_ACLAP" and "Note_control_ACLAP" column on maf output.  These two columns only contain real value when qannotate vcf2mafTmp mode is called.  To avoid confusing on the output file, we stop to output these two columns on vcf2maf mode but keep the same logic on vcf2mafTmp. 

Fixes #220 

## Type of change

# How Has This Been Tested?

related unit test are updated and created. 

# Checklist:


holmeso (approved):
LGTM

PR #223, opened 2020-12-24
Cleanup PR222

holmeso:
# Description

PR222 was merged into master before I had a chance to address the comments raised, and so this PR aims to do just that!

The main change here is to do with how `BLATRecord` objects are created.
Previously, a 21 element `String` array was passed as an argument to the constructor.
Constructing the 21 element `String` array required in-depth knowledge of the positioning of the required elements.

This PR introduces a static inner `Builder` class to the `BLATRecord` class, which explicitly defines the fields that are being set, thus alleviating the need to know where each attribute should sit in the array.

Have also created a `Builder` ctor that takes a `String` array which is used in instances where this makes more sense (ie. when constructing a `BLATRecord` from the `BLAT` class).

Have also removed the non-documenting commented out code as highlighted.

## Type of change

# How Has This Been Tested?

Unit tests have been updated to take advantage of the new Builder. Existing tests pass.
Qsv has been run against the COLO_829 GS dataset and identical results to previous runs have been obtained.

# Checklist:


delocalizer (approved):
Looks great

PR #226, opened 2021-01-28
move protocol s3 from qio to qpicard

### LINKED ISSUE #217: remove IO inrelated classes from qio to qcommon
* org.qcmg.pileup.QSnpRecord is a class to store SNP information used by qsnp and qmule project.  However, the related file reader is deprecated and only used by qmule which is not well maintained.  
* org.qcmg.protocol.s3* is URL related package only used by qpicard.  qpicard is also widely imported by most adamajava projects. At moment any project imports qpicard, automatically imports qio. It make the relationship between these projects ambiguse. 

* We move class QSnpRecord to qcommon, indicates this class is commonly imported by adamajava projects but not IO related.
* We move org.qcmg.protocol.s3*  to qcommon,  indicates this class is commonly imported by adamajava projects but not IO related.

**Additional context**
In this way the relationship between these classes and admajava projects transparent and straight.

ChristinaXu2017:
# Description

as described on #217

Fixes #217

## Type of change

# How Has This Been Tested?

existing unit test, whole gradle build and regression test

# Checklist:

### DIFF
--- a/qpicard/src/org/qcmg/picard/SAMFileReaderFactory.java
+++ b/qpicard/src/org/qcmg/picard/SAMFileReaderFactory.java
@@ -27,7 +27,7 @@ import htsjdk.samtools.ValidationStringency;
 import org.qcmg.common.log.QLogger;
 import org.qcmg.common.log.QLoggerFactory;
 import org.qcmg.common.string.StringUtils;
-import org.qcmg.protocol.s3.S3AwareURLStreamHandlerFactory;
+import org.qcmg.picard.s3.S3AwareURLStreamHandlerFactory;
 
 public class SAMFileReaderFactory {
 //	EnumSet<Option> defultOption = SamReaderFactory.Option.DEFAULTS;
--- a/qcommon/src/org/qcmg/common/model/QSnpRecord.java
+++ b/qcommon/src/org/qcmg/common/model/QSnpRecord.java
@@ -4,13 +4,9 @@
  *
  * This code is released under the terms outlined in the included LICENSE file.
  */
-package org.qcmg.pileup;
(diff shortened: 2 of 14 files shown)


holmeso:
s3 package should stay in qio as it is to do with input and output. Let me know when you have taken this out and I'll take another look - thanks

qpicard/src/org/qcmg/picard/SAMFileReaderFactory.java
```diff
@@ -27,7 +27,7 @@
 import org.qcmg.common.log.QLogger;
 import org.qcmg.common.log.QLoggerFactory;
 import org.qcmg.common.string.StringUtils;
-import org.qcmg.protocol.s3.S3AwareURLStreamHandlerFactory;
+import org.qcmg.picard.s3.S3AwareURLStreamHandlerFactory;
```
delocalizer, line 30: I really don't understand this move. The `s3` package is for low-level AWS operations - for example, all it ever throws is IOException. `picard` is specifically for handling fastq/bam/sam sequence data which is a way more specialized task. By moving this out of `qio` into `qpicard` that means any time I want to access ANY s3 file stream, whether it contains a VCF, text or whatever, I have to import a sequence data handling package. That just seems wrong to me.
A lot of my reluctance I think comes from my (very imperfect) understanding of what problem this is attempting to solve. It seems like if the issue is multiple paths to the same dependency, that is something that should be fixed in the build configuration (with excludes) rather than restructuring packages themselves.

ChristinaXu2017:
sure, I agree s3 package  is to do with IO, but qpicard is also deal with IOs. I think they are perfectly related.

PR #235, opened 2021-02-01
Qprofiler qio

ChristinaXu2017:
# Description

- The org.qcmg.ma, org.qcmg.fast and org.qcmg.gff are deleted from qio, the alternative new pacake is org.qcmg.qio.ma, org.qcmg.qio.fasta and org.qcmg.qio.gff. these new package is based on org.qcmg.qio.record.RecordReader (refer to pull request #209).
- update qprofiler by using new qio packages. 
- update qbamannotate by using new qio::ma package
- use org.junit.Assert to replace the deprecated junit.framework.Assert inside unit tests
- add unit test method to delete output qprofiler.xml after each test run. 
- tidy up code. 

## Type of change

# How Has This Been Tested?

the unit test is created and updated, regression test in our TC box was created. 

# Checklist:


qprofiler/test/org/qcmg/qprofiler/QProfilerTest.java
```diff
@@ -39,6 +40,17 @@ public void setup() throws IOException {
 		
 		createTestFile(DODGY_GFF_FILE_NAME_FILE, getDodgyFileContents());		
 	}
+	
+	@After
+	public void tidyUp() throws IOException {
```
holmeso, line 45: It's generally safer to use the `TemporaryFolder` class to setup files in each test method, as that way, junit takes care of removing the files once they have been used. I know that in this case the original code is not yours - do you know which test method is producing the `qprofiler.xml` file? If so, perhaps that method(s) can be updated to take a temp file as the output?

ChristinaXu2017: qprofiler.xml is a default output if the --output option is not called, see details on QProfiler::setup() {...if (null != options.getOutputFileName()) ...}. 
I think it is a good idea to create a default output, but it should not be always tested. The unit test is updated, we test qprofiler with --output option, unless once to test the default one exits or not.

qio/src/org/qcmg/qio/ma/MaMapping.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public String getLocation() {
+        return location;
+    }
+
+    public int getMismatchCount() {
+        return mismatchCount;
+    }
+
+//    public MAMappingParameters getParameters() {
```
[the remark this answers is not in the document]
ChristinaXu2017: done

holmeso (approved):
thanks for the changes

PR #237, opened 2021-02-04
Vcf qio

ChristinaXu2017:
# Description

- Create org.qcmg.qio.vcf.VcfFileReader based on org.qcmg.qio.record package
- Replace old VCFFilerReader with new VcfFileReader in all adamajava projects.
- Replace old VCFFileWriter with org.qcmg.qio.record.RecordWriter in all adamajava projects
- Use TemporaryFolder in q3indel and qnnotate unit tests 
- remove Deprecate annotation from qannotate::SampleColumn.java, this class seems widely used inside qannotate.

## Type of change

# How Has This Been Tested?

unit tests is updated, TC regression test will test most Adamajava tools.

# Checklist:


q3indel/test/au/edu/qimr/indel/IniFileTest.java
```diff
@@ -9,56 +9,66 @@
...
-	public void after() {	
-		fini.delete();
+	@Before
+	public void setup() throws IOException {
+		fini =  testFolder.newFile("test.ini");		
+		output = testFolder.newFile("test.output.vcf");
+		blah = testFolder.newFile("blah");	
+		
+		//must delete output the Option instance won't accept existing output
+		output.delete();
```
holmeso, line 34: I don't think you should need to create and then delete files. `TemporaryFolder` allows you to create a directory - maybe do that and have `output` and `blah` pointing to files (that don't yet exist) in the directory

q3indel/test/au/edu/qimr/indel/pileup/IndelMTTest.java
```diff
@@ -9,57 +9,62 @@
...
- 		
+	public void before() throws IOException {
+		test_bam = testFolder.newFile("test.bam");
+		inputVcf = testFolder.newFile("input.vcf");
+		ini_noquery = testFolder.newFile("test1.ini");
+		ini_query =	testFolder.newFile("test2.ini");
+		
+		//option won't accept existing file, so have to delete but record File instance
+		output = testFolder.newFile("test.output.vcf");
+		output.delete();
```
holmeso, line 45: same as before

qannotate/src/au/edu/qimr/qannotate/utils/SampleColumn.java
```diff
@@ -10,7 +10,6 @@
 import org.qcmg.common.vcf.header.VcfHeaderRecord;
 import org.qcmg.common.vcf.header.VcfHeaderUtils;
 
-@Deprecated
 public class SampleColumn {
```
holmeso, line 13: This should still be deprecated - want to encourage people to use `VcfFileMeta` instead. Its mostly test and unused classes (`VcfToMafTmp` and `CustomerConfidenceMode`) that use this. Maybe add a comment to the deprecation annotation telling users to use `VcfFileMeta` instead?

holmeso (commented):
Have you had a chance to manually test these changes on the annotation process? It would be be very comforting to know that identical output is produced (to say a production run) and that it doesn't take any longer to run.

q3indel/test/au/edu/qimr/indel/IniFileTest.java
ChristinaXu2017: I tested it, it does create a real file. eg.   output = testFolder.newFile("output.vcf"); Assert.assertTrue(output.exists());

q3indel/test/au/edu/qimr/indel/pileup/IndelMTTest.java
ChristinaXu2017: same as above

q3indel/test/au/edu/qimr/indel/IniFileTest.java
holmeso: but why are you then immediately deleting it?

qannotate/src/au/edu/qimr/qannotate/utils/SampleColumn.java
ChristinaXu2017: Thanks for explaiantaion, I put the "Deprecated" back, and will add your suggested comment on next pull request

q3indel/test/au/edu/qimr/indel/IniFileTest.java
ChristinaXu2017: It only needs the file path. If the output file already exists, then q3indel will throw an exception.

holmeso: rather than creating and then deleting a file - why not create a directory for the output (as mentioned above)?

ChristinaXu2017: OK, did and pushed

ChristinaXu2017:
> Have you had a chance to manually test these changes on the annotation process? It would be very comforting to know that identical output is produced (to say a production run) and that it doesn't take any longer to run.

failed on qsv call. I am investigating now

ChristinaXu2017:
I pushed my local build jars to /software/adamajava/adamajava-release/xuTest_adamaJavaVersion, and then run somaticDnaFastqToMaf with jars inside this folder.  It was failed during qsvControlVsTest. 
hence I run the qsv again with our nightly version and xuTest_adamaJavaVersion. But both suceed, and the results passed by our regression_GS_latest_bams.sh; regression_GS_latest_mafs.sh and regression_GS_latest_xmls.sh

ChristinaXu2017:
In this pull request, there is no change in qsv. there is the only qsv that failed during our manual test and then succeed during re-run.  It seems there is a random exception inside qsv but nothing to do with this pull request.

ChristinaXu2017:
after 10 hours run, the WGS failed due to a permission issue.  I have to change root dir to temp/mock_analysis and re-run it.

ChristinaXu2017:
Cromwell job dye again due to the walltime exceed

ChristinaXu2017:
It seems there are different WDLs for somaticFastqtoMaf.  The one in our genome info svn may not cope with the WGS. 
As a developer, it is impossible to run on such a big data set. We can only run on a small dataset with limited resources.  The best way to test it is as an expert to run it before release a new version (tag it). 
This is why we have merge, release, adama-nightly, adama-current.

holmeso:
> It seems there are different WDLs for somaticFastqtoMaf. The one in our genome info svn may not cope with the WGS.

The wdl should work against WGS datasets - thats what is used in production currently. Perhaps you chose the wrong wdl?
Would it not be simpler to just run the annotation process against bams that already exist? Running the whole wdl seems overkill to just test the reading/writing of the vcf files.

> As a developer, it is impossible to run on such a big data set. We can only run on a small dataset with limited resources. The best way to test it is to ask an expert to run it before release (tag it).

I'm not sure I agree with this - as developers we need to ensure that the code we have written has been tested against data as close to production level data as possible. If this is not possible for developers to do, then I'm not sure how feasible it's going to be for other people to do..

ChristinaXu2017:
There are several copies of this WDL pipeline, The one I pick up from genome svn; but our production team uses the one from a different location, eg. /working/lab_nic.  they are slightly different, eg. the parameter settings inside qprofiler2.wdl are different. 

This pipeline is not only calls our adamajava, we are not the right people to trouble shotting this pipeline.  What we should do is only to test our adamajava. eg as qannotate, q3indel, qprofiler etc.  I have to give up to run the somaticFastqtoMaf on a WDL dataset.

ChristinaXu2017:
It is dated, no value to merge to master

PR #251, opened 2021-03-08
set default header discriminator hash

ChristinaXu2017:
# Description

The default file header discriminator is different between `StringFileReader` and its parent class `RecordReader`.  Constructor `StringFileReader(File file, int bufferSize)` uses the default header discriminator from the parent class, which is null. Here we fix it by changing it to `HASH_STRING`. 
 
## Type of change

# How Has This Been Tested?

new unit tests added.

# Checklist:


holmeso (approved):
looks good

PR #254, opened 2021-03-24
qsignature - allow snpPositions file to have headers

holmeso:
# Description

`qsignature` fails if the snp positions file contains a header.
This has been fine in the past, because we only ever had a single snp positions file.
There is now some work being done to create a GRCh38 version of the snp positions file, and having a header in the file would be very useful.

## Type of change

# How Has This Been Tested?

Additional unit tests have been added to test that snp positions files with headers (as well as those without) can be loaded correctly
Additionally, the project was build and run against a production WGS bam file using a snp positions file with headers successfully.

# Checklist:

### DIFF
--- a/qsignature/src/org/qcmg/sig/SignatureGeneratorBespoke.java
+++ b/qsignature/src/org/qcmg/sig/SignatureGeneratorBespoke.java
@@ -278,7 +278,7 @@ public class SignatureGeneratorBespoke {
 			}
 			
 			/*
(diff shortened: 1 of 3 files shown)


grendeloz (approved):
Looks good to me.  There is a small typo - "illumia" is used instead of "illumina" in a few places but we can fix that at some future stage and it doesn't affect functionality obviously.

PR #255, opened 2021-03-24
document update for qmotif

ChristinaXu2017:
# Description

Updating documentation (readthedocs.io) and messages.properties file. Hence the tool usage message while using the command "java -jar qmotif.jar -h", matches reading .readthedocs.io.  Option related java code is also updated, the help usage shows as below. 

```
usage: java -jar qmotif.jar --ini <ini_file> --input-bam <input_bam_file> --output-xml <xml_file> --output-bam <bam_file> --log <logfile> [options]
Option               Description                            
------               -----------                            
-h, --help           Show usage and help.                   
--ini                Req, Configuration file in INI format. 
                       This file is designed to contain     
                       parameters that are likely to be     
                       fixed for a given type of analysis.  
                       So the general idea is that you test 
                       parameters in the INI file until you 
                       are happy and then you use that      
                       single INI file with all of the BAM  
                       files that are part of this analysis.
--input-bai          Req, Index file for the BAM file being 
                       processed.                           
--input-bam          Req, BAM file on which to operate.     
--log                Req, Log file.                         
--loglevel           Opt, Logging level [INFO,DEBUG], Def=INFO.                            
--output-bam         Req, Output a BAM file containing all  
                       reads that matched the motif search. 
--output-xml         Req, Output XML file containing the  motifs
                       found and the regions they were found in.                       
--query              Req, Query string for selecting reads for coverage.                        
--threads <Integer>  Opt,Number of threads to use when running qmotif. Def=1.               
--validation         Opt, BAM record validation stringency  
                       [STRICT,LENIENT,SILENT]. Def=LENIENT.
--version            Show version number.          
```

## Type of change
Documentation

# How Has This Been Tested?

- unit test is updated, 
- command-line is tested, the result as above. 
- the updated wdl pipeline was run twice.  The first one call our production adamajava qmotif, the second one call local build qmotif with new options. 
 -- call production, wdl: {"id":"26451940-ae23-4532-8df1-501d1ea6a694","status":"Submitted"}  
 -- call new qmotif, wdl: {"id":"6d4e403d-913d-41dc-8063-592b3fdc66ad","status":"Submitted"}
 -- both output BAM file have same md5 string 
 -- both XML outputs are identical except the version and wdl directory

# Are WDL Updates Required?

both imports/calling/qmotif.wdl and workflow/analysis/qmotif.wdl are updated. There is a new task which calls qmotif with new options. 
 
# Checklist:


qmotif/src/org/qcmg/motif/Summariser.java
```diff
@@ -26,6 +26,7 @@
 import org.qcmg.common.string.StringUtils;
 import org.qcmg.common.util.LoadReferencedClasses;
 
+@Deprecated
```
holmeso, line 29: Just curious as to why this class (and its test ) is being deprecated?

ChristinaXu2017: This class is not used anywhere, also can't find any related information in document.  It seems replaced by org.qcmg.motif.Motif,  class Motif is the mainClassName.

holmeso: > This class is not used anywhere, also can't find any related information in document. It seems replaced by org.qcmg.motif.Motif, class Motif is the mainClassName.

It's used to `Summarise` the `qmotif` xml files. The `Motif` class is performing a different function

holmeso:
What changes to any WDL files are required as part of this change?

holmeso:
Have you compared the output of your changes against output from the existing code run against production data to check that they are the same?

qmotif/src/org/qcmg/motif/Summariser.java
ChristinaXu2017: sure, I can remove de deprecated annotation if it is still used. However, could you create a related usage or document?

ChristinaXu2017:
good suggestion, I think I have to compare the output against the change. However I can't find any existing data run by qmotif, I don't think they belong to WDL pipeline.  But I can compare this one to the production one.

ChristinaXu2017:
the comparision of my changes against production qmotif is done, the outpus is /mnt/lustre/working/genomeinfo/share/testdataXu/adama.doc/qmotif

holmeso:
Looks like there are some related wdl files:
` production/wdl/imports/calling/qmotif.wdl`
and
`production/wdl/workflow/analysis/qmotif.wdl`
Be good to know what changes to these files will be required once this change has been released

holmeso:
> Looks like there are some related wdl files:
> ` production/wdl/imports/calling/qmotif.wdl`
> and
> `production/wdl/workflow/analysis/qmotif.wdl`
> Be good to know what changes to these files will be required once this change has been released

Hi Christina, have you had a chance to look at this? The locations will have changed due to the recent trunk/tag/branch upgrade

ChristinaXu2017:
I updated both production/wdl/trunk/imports/calling/qmotif.wdl and production/wdl/trunk/workflow/analysis/qmotif.wdl
A new task called qmotif.current run the qmotif from this pull request, and previous qmotif.qmotif run the qmotif from master.  An if condition added into the analysis/motif.wdl, it can call both tasks. 

I know this both WDL have to update again after merge.

holmeso:
Could `Req` be expanded to `Required`, and `Opt` to `Optional` in the help usage description fields please?
That means there is no possibility of mis-understanding the options

holmeso:
> please refer to JP's note svn/genomeinfo/production/mkdocs/docs/tutorials/stuff_to_do.md, We will use Req and Opt in all our adamjava tools

If that's what JP wants, that's what we'll do!

ChristinaXu2017:
> Would it not be better to make the changes in production/wdl/branch and then merge branch with trunk once a new adamajava version has been released?

sounds good, but at moment not sure which branch we should use. all our regression test is based on trunk, there is no links between Adama release and branch/trunk, not sure how to follow it.

holmeso (approved):

PR #256, opened 2021-03-31
bug fix in toString method of VcfRecord

holmeso:
# Description

There is currently a bug in the `VcfRecord.toString` method that adds an extra tab when the record does not have any format fields.
eg.
If the following VcfRecord has toString called on it:
```
chr1\t1000\trs100\tA\tC\t0.0\tPASS\t.
```
it ends up looking like:
```
chr1\t1000\trs100\tA\tC\t0.0\tPASS\t.\t\n
```
As you can see, there is an extra tab before the new line character. 
This will cause the `VcfRecord` to fail validation

## Type of change

# How Has This Been Tested?

Added unit test, modified existing unit test.

# Checklist:


ChristinaXu2017 (approved):
Looks good

PR #259, opened 2021-04-21
Update pull_request_template.md

holmeso:
Updating the PR template so that it includes a section on whether any WDL needs to be updated.


ChristinaXu2017 (commented):
not sure whether qmotif is a good example for wdl update. It seems qmotif not used by our WDL pipeline. I guess qprofiler maybe a good example here

holmeso:
> not sure whether qmotif is a good example for wdl update. It seems qmotif not used by our WDL pipeline. I guess qprofiler maybe a good example here

qmotif was just an example - feel free to change it to another project if you would like.

ChristinaXu2017 (approved):
agree

PR #263, opened 2021-05-18
Qsv.doc

ChristinaXu2017:
# Description

1. Tool options are updated and new usages show as below:
> -  log, log level options are removed from the command line. Because Qsv only picks up log file name from ini file, and output to a random folder for each run. it is out of command-line control.
>   -  the range option is removed from the command line. it overlaps with the INI file and confuses the user. 
>   -   the "tmp" option renamed to "output-temporary"; "overrideOutput" renameed to "output". It follows our universal Adamajava option rule as suggested by JP.
> 

```
usage: java -jar qsv.jar --ini <ini_file> --output-temporary <directory> [OPTIONS]
Option              Description                                                           
------              -----------                                                           
--help              Show usage and help.                                                    
--ini               Req, ini file with required options.                                    
--output            Opt, specify qsv output directory here, the base name must be uuid      
                      string. All existing output will be override.  Without this option qsv
                      will pick up "output" option from ini file.                           
--output-temporary  Req, directory to write temporary files to.                             
--version           Show version number.                 
```

2.  ini file options and descriptions are also updated, details and template is listed in README.md. 

>   -   BLAT information are removed due to BLAT class is deprecated, now we use our in-home aligner - q3tiledaligner;
>   -    Option of isize_records under [general] and primer_size under [pair]  are removed. They was dated and no longer used inside qsv;
>   -    New descriptions are added for ini option : "general::qcmg", "general::tiled_aligner", "test::sample_id", "control::sample_id"

3.  unused dependency "qbamannotate"  is deleted from gradle build file. 

## Type of change

# How Has This Been Tested?

The existing unit passed, new options are tested.  The command-line help usage is shown on above. 
We run it on real data and compare it to TC box result #326 which tool version is 66-9be730f8.  The qsv run with ToolVersion 66-9be730f8 was successful by using tiled_aligner but not Blat.  Both runs got the same counts (*.qsv.summary.txt and *_sv_counts.txt)

# Are WDL Updates Required?

qsv.wdl has to be changed before we use "--tmp", "--overrideOutput" now we use "--output-temporary", "--output". qsv_ini.wdl may require an update in soon future, this blat related information is no longer used in qsv, however qsv require a further update, at moment it still reads blat information from ini file but the value is not further used. 

# Checklist:


grendeloz:
The description says that --output must be a uuid string.  Why is this?  I can imaging that in our use it would be true but are we rally forcing this on all users?

grendeloz (commented):
The documentation in the README file should all be moved into the docs tree.

As a rule, individual tools should not have README files at all.  The main project README file points to the ReadTheDocs which is where we should be directing people.

Could you please remove the README files for qmotif, qannotate, qsv and qprofiler.  If there are any changes in there, move them to the docs tree.

ChristinaXu2017:
As we discussed the doc tree will be separated from our Adamajava, an independent repo will be created.  At moment most Readme are overlap with doc three, they should be deleted, or replace the old files under docs tree.

Temporary I put ini options inside the qsv readme. I will create a new doc for our updated qsv, but only after doc repo is ready.

ChristinaXu2017:
> The description says that --output must be a uuid string. Why is this? I can imaging that in our use it would be true but are we rally forcing this on all users?

Each Qsv run will create an analysis id in UUID format, which will be the prefix for all output names including the output folder.  I think it rather says "analysis id" than "output folder name".  However this option takes the file path but not a pure uuid string, so it is point to an output location.

PR #264, opened 2021-06-08
Qprfiler2.doc

ChristinaXu2017:
# Description

Tool options are updated, it follows our universal adamajava option rule as suggested by JP. The help usage shows below. 

```
usage: java -jar qprofiler2.jar --input <input file> --output <output xml file> --log <log file> [options]
Option                        Description                                                               
------                        -----------                                                               
--bam-full-header             Opt, this option indicate qprofiler2 output entire BAM header ; otherwise 
                                only output HD and SQ.                                                  
--bam-index                   Opt, a BAI index file. The option of "bam-threads-producer" will be       
                                ignored if no index file provided.                                      
--bam-records <Integer>       Opt, only process the specified number of records from the beginning of   
                                the BAM file. Default to process entire BAM file.                       
--bam-validation              Opt, BAM record validation stringency [STRICT,LENIENT,SILENT].            
                                Def=LENIENT.                                                            
--help                        Show usage and help.                                                      
--input                       Req, input file in FASTQ, BAM/SAM or VCF format, also support FA, MA,     
                                QUAL, FASTA, GFF and GFF3 format.                                       
--log                         Req, Log file.                                                            
--loglevel                    Opt, Logging level [INFO,DEBUG], Def=INFO.                                
--output                      Opt, XML output which containing basic summary statistics. Def="qprofiler.
                                xml".                                                                   
--threads-consumer <Integer>  Opt, number of threads to process inputed records, BAM and FASTQ mode     
                                only. Def=0 (not multi-thread mode).                                    
--threads-producer <Integer>  Opt, number of threads to read the indexed BAM file, BAM mode only. This  
                                option will be ignored without "threads-consumer" option.  Def=1.       
--vcf-format-field            Opt, group VCF records according to user specified format fields. eg.     
                                "GD", "CCM","FT=PASS". Default to ignore VCF format field.              
--version                     Show version number.                                           
```

Fixes # (issue)

## Type of change

# How Has This Been Tested?

The existing unit passed, new options are tested.  The command-line help usage is shown above. 
Test on real BAM from  Team city: #351 (11 Jun 21 01:30)
compare xml from current qprofiler2 with  a released qprofler2(ToolVersion 39-b903c82c). Got same md5 string, bash script is `sed 's/>/>\n/g' $xml | sed -n '/<bamSummary/,/bamSummary>/p' | sed -n '/<readGroup name="fa428f11-7798-4cf6-8519-39d4dac95f04"/,/readGroup>/p' | md5sum`
 

# Are WDL Updates Required?

qprofiler2.wdl has to be changed, before we use "--ntConsumer" "--ntProducer" now we use "--threads-consumer", "--bam-threads-producer";  the option "--maxRecords" is also changed to "--bam-records", but it is mainly used internal, won't affect any current WDL. 

# Checklist:


grendeloz:
For consistency, I think we should drop the -tc and -tp short forms.

Also, qprofiler operates on many file types BUT from the descriptions, it looks like quite a few of the options ONLY relate to BAM files.  Is is true that these options are only for BAM files:

1. --full-bam-header
2. --records
3. --threads-consumer
4. --threads-producer
5. --validation

If yes, should all of these options have a bam prefix to make that clear, e.g.:

1. --bam-full-header
2. --bam-records
3. --bam-threads-consumer
4. --bam-threads-producer
5. --bam-validation

 Thoughts?

ChristinaXu2017:
got it

holmeso:
Are we happy with the option names `--bam-threads-producer` and `--threads-consumer` ?
I get that the reason for this is due to `threads-producer` being used by more than the `bam` summariser, but it may be a little confusing for users.
Having said that, if its documented, then maybe it isn't an issue

holmeso (approved):

PR #267, opened 2021-06-15
Qsv.option

ChristinaXu2017:
# Description

1. Tool options are updated and new usages show as below:
> -  log, log level options are removed from the command line. Because Qsv only picks up log file name from ini file, and outputs to a random folder for each run. it is out of command-line control.
>   -  the range option is removed from the command line. it overlaps with the INI file and confuses the user. 
>   -   the "tmp" option renamed to "output-temporary";  It follows our universal Adamajava option rule as suggested by JP.
>   -   the "overrideOutput" renamed to "uuid".  Qsv requires an UUID as analysis id, a sub-folder named by this UUID will be created under the output directory specified in the INI file. All Qsv outputs will be stored under this folder, all existing data will be overridden if it already exists. By default, A random UUID will be created during each Qsv run.  Previously, option "overrideOutput" require an output directory, it conflicts with the "output" option specified inside INI file; it also requires the basename must be random UUID formate, which also confuses the user. 

```
usage: java -jar qsv.jar --ini <ini_file> --output-temporary <directory> [OPTIONS]
Option              Description                                                           
------              -----------                                                           
--help              Show usage and help.                                                    
--ini               Req, ini file with required options.                                    
--uuid              Opt, a subfolder under output directory will be named by this customized
                      UUID. By default, a random UUID will be created.                         
--output-temporary  Req, directory to write temporary files to.                             
--version           Show version number.                 
```

2.  INI file options and descriptions are also updated, details and template are listed in README.md; this README.md is moved to adamajava/docs and renamed as qsv.latest.md

>   -   BLAT information are removed due to BLAT class is deprecated, now we use our in-home aligner - q3tiledaligner;
>   -    Option of isize_records under [general] and primer_size under [pair]  are removed. They was dated and no longer used inside qsv;
>   -    New descriptions are added for ini option : "general::qcmg", "general::tiled_aligner", "test::sample_id", "control::sample_id"

3.  unused dependency "qbamannotate"  is deleted from gradle build file. 

## Type of change

# How Has This Been Tested?

The existing unit passed, new options are tested.  The command-line help usage is shown on above. 
We run it on real data and compare it to TC box result #326 which tool version is 66-9be730f8.  The qsv run with ToolVersion 66-9be730f8 was successful by using tiled_aligner but not Blat.  Both runs got the same counts (*.qsv.summary.txt and *_sv_counts.txt)

We also run the new Qsv with and without new option "uuid"; it functioned as what we desired: 

- create a sub folder name by a randome UUID and Qsv will output to this folder, without option "uuid".
- create or override the subfolder named by the user specified UUID,   with option "uuid".

# Are WDL Updates Required?

qsv.wdl has to be changed before we use "--tmp", "--overrideOutput" now we use "--output-temporary", "--output". qsv_ini.wdl may require an update in soon future, this blat related information is no longer used in qsv, however qsv require a further update, at moment it still reads blat information from ini file but the value is not further used. 

# Checklist:


holmeso (approved):
looks good

PR #269, opened 2021-06-28
remove qprofiler/README to docs/qprofiler

ChristinaXu2017:
# Description

Tidy up documents of Qprofiler:

- Remove README file from Qprofiler folder to docs/qprofiler, and rename as qprofiler_V1.md.
- The qprofiler_2_0.md lists qprofiler-2.0.jar  from qprofiler2.0.tar.bz2, but they do not exist under our Adamajava repository.  Here  qprofiler_2_0.md is replaced by qprofiler_V1.md.
- update qprofiler_1_0.md and rename it to qprofiler_V1.md. There are two jars (qprofiler-1.0.jar and qprofiler.jar) in Adamajava release, but both contain the same option and functions.  The 'qprofiler_V1.md' lists all these jars' usage.
- minor modify some options: delete short opt "--tc" and "--tp"; mark "--include" as deprecated. 

## Type of change

- [x] Breaking change (two short option deleted, but they are not widely used.)
- [x] This change requires a documentation update

# How Has This Been Tested?
All documents update no tests required.  The option modification is unit tested

# Are WDL Updates Required?

Current changes won't affect WDL.

# Checklist:

n/a


holmeso:
Can I confirm that when you are talking about qprofiler_V2 here, it is a version of the original qprofiler rather than the qprofiler2 project?

ChristinaXu2017:
> Can I confirm that when you are talking about qprofiler_V2 here, it is a version of the original qprofiler rather than the qprofiler2 project?

ChristinaXu2017:
The usage options are changed, so we have to make a new document for qprofiler. It is not the qprofiler2.

holmeso (approved):

PR #272, opened 2021-07-16
Qsig remove tokenising

holmeso:
# Description

A couple of changes to support qsignature's foray into the world of different (larger) positions files.
First, some instances where the `TabTokenizer` class was being used unnecessarily have been replaced with a less cpu intensive process (according to jvisualise)

And secondly, a new option (maxCacheSize) has been introduced that is used by the `Compare` class.
This option allows the user to specify the number of qsignature vcf files that should be stored in cache for the comparison.
Traditionally, all files would be stored in cache and then the comparisons performed, but when positions files are large, and therefore qsignature vcf files are also large, this method has the potential to cause `OutOfMemory` errors.

The downside to limiting the cache size is that it means that some files will be loaded more than once, pushing out the runtime as well as increasing the I/O impact.

## Type of change

# How Has This Been Tested?

Unit tests have been expended.
Updated code has been run and compared against existing code with identical results.

# Are WDL Updates Required?

No wdl updates are required. The default behaviour of the new option is to proceed as it would have previously.

# Checklist:


qsignature/test/org/qcmg/sig/CompareTest.java
```diff
@@ -160,6 +160,64 @@ public void bespokeVsBespokeBAM() throws IOException, InterruptedException {
...
+		writeDataToFile(bespoke2Data, bespoke2);
+		
+		Executor exec = execute("--log " + logF.getAbsolutePath() + " -d " + bespoke1.getParent() + " -o " + o.getAbsolutePath() + " --max-cache-size 0");
+		assertEquals(0, exec.getErrCode());		// all ok
+		assertEquals(true, o.exists());
+		List<String> outputData = Files.readAllLines(Paths.get(o.getAbsolutePath()));
+		assertEquals(10, outputData.size());		// 10 lines means 1 comparison
+		assertEquals(true, outputData.contains("<comparison file1=\"1\" file2=\"2\" overlap=\"5\" score=\"1.0\"/>"));
+		
+		exec = execute("--log " + logF.getAbsolutePath() + " -d " + bespoke1.getParent() + " -o " + o.getAbsolutePath() + " --max-cache-size 1");
```
ChristinaXu2017, line 192: All these five Executors are same results, why not just loop rather than repeat five times

qsignature/src/org/qcmg/sig/messages.properties
```diff
@@ -32,6 +32,7 @@ UNKNOWN_ERROR = An unknown error has occurred
 INPUT_FILE_READ_ERROR = Could not read input file {0}
 MIN_COVERAGE_OPTION_DESCRIPTION = Specifies the minimum coverage to be used in the comparison
 CUTOFF_OPTION_DESCRIPTION = Specifies the cutoff value to be used in the comparison
+MAX_CACHE_SIZE_OPTION_DESCRIPTION = Opt, number of files to store in the cache. Used by the Compare class. Set this when the qsignature vcf files to be compared have a large number of positions. Defaults to -1 (unlimited number of files in cache)
```
ChristinaXu2017, line 35: It is great to put "Opt" at the begin, but could you use "Def=-1" as JP suggested.

qsignature/src/org/qcmg/sig/Options.java
```diff
@@ -93,6 +96,8 @@
 		.describedAs("hetLowerCutoff");
 		parser.accepts("noOfThreads", NO_OF_THREADS_OPTION_DESCRIPTION).withRequiredArg().ofType(Integer.class)
```
ChristinaXu2017, line 97: This Long-form option name are not followed JP's docs rule. Here I think the option "--threads" is enough to indicate "number of threads".

qsignature/src/org/qcmg/sig/Options.java
```diff
@@ -141,6 +146,7 @@
 		minMappingQuality = ((Integer) options.valueOf("minMappingQuality"));
 		minBaseQuality = ((Integer) options.valueOf("minBaseQuality"));
 		noOfThreads = ((Integer) options.valueOf("noOfThreads"));
+		maxCacheSize = ((Integer) options.valueOf("maxCacheSize"));
```
ChristinaXu2017, line 149: "maxCacheSize" doesn't follow JP's docs rule, it also confuse by the word "size". How about something like  "--cached-vcfs", it means how many qsigniture VCF files can be stored in RAM.

ChristinaXu2017 (commented):
as describe inside code

qsignature/test/org/qcmg/sig/CompareTest.java
holmeso: The `max cache size` option is different for each of the executions

qsignature/src/org/qcmg/sig/messages.properties
holmeso: Sure

qsignature/src/org/qcmg/sig/Options.java
holmeso: Out of scope for this PR

qsignature/src/org/qcmg/sig/Options.java
holmeso: I provided both the maxCacheSize and max-cache-size options for this property as a bridge between the 2 worlds. I didn't want to just put in the new way as no other option for this tool follows that structure and so both were put in with the view of removing maxCacheSize when the other options are updated (out of scope for this PR)

ChristinaXu2017 (approved):
Looks good

PR #291, opened 2021-12-20
Qsignature - use cutoffs specified in options

holmeso:
# Description

The `Compare` class within `qsignature` needs to make decisions on whether a snp is hom or het. It uses 3 values to determine this: 

- hom cutoff (defaults to 0.9)
- lower het cutoff (defaults to 0.3)
- upper het cutoff (defaults to 0.7)

There are 3 existing options that allow the user to override these defaults, however, the `Compare` class does not take these into account.

This PR rectifies this oversight, allowing the user to set the cutoffs as they see fit.

A record of the cutoffs used is now added to the xml output so that the viewer can tell what cutoffs were used.

## Type of change

# How Has This Been Tested?

Existing unit tests have been updated and new tests added.

# Are WDL Updates Required?

No wdl updates are required.
It is anticipated that once this change is in master, an update to the `SDFTM` wdl workflow will be made to supply cutoff values to `qsignature`.

# Checklist:


ChristinaXu2017 (approved):
Looks good

PR #293, opened 2022-01-07
remove class loading

### LINKED ISSUE #213: class loading issue
**Describe the bug**
Got below error in adamaJava 36-ef3dc161
`$java -jar q3panel.jar -h
java.lang.SecurityException: sealing violation: package nu.xom is sealed`
and 
`$java -jar q3tiledaligner.jar -h
Exception in thread "main" java.lang.IncompatibleClassChangeError: class org.apache.bcel.util.ClassPath$2 has interface org.apache.bcel.util.ClassPath$ClassFile as super class
`
**Reason**
Actually, the reason was of having a duplicate library, one is located within the classpath and the other one is included inside a jar file that is located within the classpath.  It's generally the result of two mildly different versions of the same dependency on the classpath and is almost always caused by either a different superclass than was originally compiled against being on the classpath or some import of the transitive closure being different, but generally at class instantiation and constructor invocation. (After successful class loading and ctor invocation, you'll get NoSuchMethodException or whatnot.) please refer to [this link](https://stackoverflow.com/questions/1980452/what-causes-java-lang-incompatibleclasschangeerror/1980474#1980474).

Both q3panel and q3tiledaligner depend on qmule and qcommon, but qmule depend on qio, qpicard and qcommon; Hence qcommon are loaded twice. 

**Solution1**
step 1: move commonly used class from qmule::SmithWatermanGotoh to qcommon::model::SmithWatermanGotoh
step 2: change dependency from 
```...
dependencies {
    ...
    api project(':qcommon')
    api project(':qmule')`
    ...
```
to 
```...
dependencies {
    ...
    api project(':qcommon')
    api project(':qio')
    ...
```

**Solution2**
stop loading java classes during runtime dynamically to the JVM.
```	
public static void main(String[] args) throws Exception {
	// loads all classes in referenced jars into memory to avoid nightly build sheninegans
	//LoadReferencedClasses.loadClasses(Q3Panel.class);
```

**Desktop (please complete the following information):**
 - OS: OS: linux or Mac
 - Version 3.10.0-693.11.6.el7.x86_64 or mac10.14.6

holmeso:
# Description

In days of old, we would execute long-running processes against jars that were overwritten nightly.
The effect of re-writing the jar files as a process was running would invariably kill the process.
The solution was to pre-load all the classes in the package and in the dependent jar files up front rather than when required.
This was done using the `LoadReferencedClasses.loadClasses` method.
This allowed the process to continue despite the jar files being re-written.

I am happy to say that this method of running against an often updating jar is no longer common place and so this method should not be used as a matter of course.
It remains in place (and un-deprecated) as there may arise instances where its utility may be useful.

All classes (apart from the test class) that called this method have been updated to remove the call to this method.

Fixes #213

### DIFF
--- a/q3indel/src/au/edu/qimr/indel/Main.java
+++ b/q3indel/src/au/edu/qimr/indel/Main.java
@@ -22,7 +22,6 @@ public class Main {
 			System.exit( 0 );
 		} 
 
-		LoadReferencedClasses.loadClasses(Main.class);
 		QLogger logger = options.getLogger();
 		int exitStatus = 0;
 		try {
--- a/q3panel/src/au/edu/qimr/panel/Q3ClinVar.java
+++ b/q3panel/src/au/edu/qimr/panel/Q3ClinVar.java
@@ -1549,8 +1549,6 @@ public class Q3ClinVar {
 	
 	
 	public static void main(String[] args) throws Exception {
-		// loads all classes in referenced jars into memory to avoid nightly build sheninegans
-		LoadReferencedClasses.loadClasses(Q3ClinVar.class);
 		
 		Q3ClinVar qp = new Q3ClinVar();
 		int exitStatus = qp.setup(args);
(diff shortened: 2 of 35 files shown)


ChristinaXu2017 (approved):
totally agree

PR #313, opened 2022-03-16
fix(q3tiledaligner): remove dependence on GRCh37

holmeso:
# Description

The `PositionChrPositionMap` class was hard-coded to use the GRCh37 reference genome. This has been
changed so that any reference index file can be used.
A check to see if the generated map contains any overlapping regions has been added.

re #307

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Additional unit tests have been added. Existing unit tests have been modified. All tests pass.

# Are WDL Updates Required?

No

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

PR #314, opened 2022-03-29
fix(qsv): incorrect index value was being passed to method

holmeso:
# Description

In `TiledAlignerUtil.loadReferenceIndexMap` a check is performed on the supplied index file. If this is null, an attempt to locate the index file is then done based on the reference file. If this is found then that should be used.
Unfortunately the code was using the supplied index value (which was null) rather than the located index file (not null).

This was introduced in PR #313 

## Type of change

# How Has This Been Tested?

New test added, existing tests pass

# Are WDL Updates Required?

No

# Checklist:

### DIFF
--- a/q3tiledaligner/src/au/edu/qimr/tiledaligner/util/TiledAlignerUtil.java
+++ b/q3tiledaligner/src/au/edu/qimr/tiledaligner/util/TiledAlignerUtil.java
@@ -1336,7 +1336,7 @@ public class TiledAlignerUtil {
 		return nonOverlappingRecs;
 	}
 
-	public static Map<ChrPosition, LongRange> loadReferenceIndexMap(String refFile, String refIndexFile) {
+	static Map<ChrPosition, LongRange> loadReferenceIndexMap(String refFile, String refIndexFile) {
 		Map<ChrPosition, LongRange> headerMap = null;
 		if (null == refIndexFile) {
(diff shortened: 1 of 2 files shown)


q3tiledaligner/src/au/edu/qimr/tiledaligner/util/TiledAlignerUtil.java
```diff
@@ -1336,7 +1336,7 @@ public static boolean doesSequenceHaveMostlySingleBaseRepeats(String sequence) {
 		return nonOverlappingRecs;
 	}
 
-	private static Map<ChrPosition, LongRange> loadReferenceIndexMap(String refFile, String refIndexFile) {
+	public static Map<ChrPosition, LongRange> loadReferenceIndexMap(String refFile, String refIndexFile) {
```
ChristinaXu2017, line 1339: It seems refIndexFile is not used in this function,  should it change to public static Map<ChrPosition, LongRange> loadReferenceIndexMap(String refFile) ?

holmeso: `refIndexFile` is used (near the end of the method).
Access changed from public to package protected

ChristinaXu2017 (approved):
Got it, looks good!

PR #316, opened 2022-04-13
refactor(q3tiled_aligner): allow user to pass in cached ref index map

holmeso:
# Description

Allow user to pass in cached ref index map. This will greatly reduce I/O.

## Type of change

# How Has This Been Tested?

Existing tests pass. New test introduced.

# Are WDL Updates Required?

n/a

# Checklist:

### DIFF
--- a/q3tiledaligner/src/au/edu/qimr/tiledaligner/util/TiledAlignerUtil.java
+++ b/q3tiledaligner/src/au/edu/qimr/tiledaligner/util/TiledAlignerUtil.java
@@ -748,18 +748,49 @@ public class TiledAlignerUtil {
 		});
 	}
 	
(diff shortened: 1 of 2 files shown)


ChristinaXu2017:
It is good to move the method loadReferenceIndexMap(String refFile, String refIndexFile) from the sub-method to the top-level method. So reduced the frequency to load the index file.   It seems "String refFile" is no longer used in new method getBlatRecordsSWAll(...)

 I suggest to change 
List<BLATRecord> getBlatRecordsSWAll(String refFile, String refIndexFile, TIntObjectMap<int[]> cache, String sequence, final String name, int tileLength, String originatingMethod, boolean log, boolean recordsMustComeFromChrInName)  
to 
public static List<BLATRecord> getBlatRecordsSWAll( Map<ChrPosition, LongRange> refIndexMap, TIntObjectMap<int[]> cache, String sequence, final String name, int tileLength, String originatingMethod, boolean log, boolean recordsMustComeFromChrInName)

ChristinaXu2017 (approved):
sorry, my bad.  I agree the reference file has to stay inside the new method.

PR #320, opened 2022-05-10
update MODE:add header line in remove mode

ChristinaXu2017:
# Description

qpileup remove mode removes the BAMs from the pileup collection, and adds a header line: "##METADATA=MODE:remove,...".  However, the "##METADATA=MODE:add" line point to that BAM is still reminded in the HDF header.  It shouldn't be a problem, when both "remove" and "add" header lines exist.  While, in merge mode, it throws away the "remove" header line, but only keeps the "add" header line.  Hence problem occurs, the merged HDF header provides incorrect information. 

Here, we delete the the "##METADATA=MODE:add" header line when a matched BAM is removed. So the correct information can pass to the downstream pipeline.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

It is too expensive to create unit test for the updated method, here we used a bash script to run qpileup and check the final output hdf file header. The test is passed

# Are WDL Updates Required?

not required

# Checklist:


holmeso (approved):
LGTM

PR #321, opened 2022-06-27
create tool to rescue fastq reads with empty sequence

ChristinaXu2017:
# Description

A Fastq file may contain an empty sequence after cutAdapter,  Here A base N is set to an empty sequence with 0 base quality value". This tool takes fastq file as input and output, but also can read from stdin and writer to stdout, so it can insert into the pipe on our new WDL script. 

## Type of change

# How Has This Been Tested?

unit tests is added, tested on small real data. It will also run by new WDL

# Are WDL Updates Required?
It is an independent new took, and not yet used by any current WDL pipeline.   Any current WDL pipeline won't be affected.  A new WDL script is created, which implements this tool after cutAdapter, and pass output fastq to BWA mapping, and also creates an uBAM (unmapped BAM).  Thi new WDL script will be tested after its branch is merged into the master.  This new WDL script named fasqTobam.wdl may replace the current mapPairedChunks.wdl.

 

# Checklist:

### DIFF
--- a/qmule/src/org/qcmg/qmule/FixCutAdapter.java
+++ b/qmule/src/org/qcmg/qmule/FixCutAdapter.java
@@ -20,7 +20,7 @@ import java.io.InputStreamReader;
 import java.util.Date;
 
 public class FixCutAdapter extends CommandLineProgram {
-    static final String USAGE = "When cutadapt finds adapter sequence at the start of a read the option --trim-n results in empty sequence in the output. This tool replaces any empty sequence with a single base N with 0 base quality value.";    
+	static final String USAGE = "When cutadapt finds adapter sequence at the start of a read the default action of adapter sequence trimming results in empty sequence in the output. This tool replaces any empty sequence with a single base N with 0 base quality value.";
     @Argument(shortName = StandardOptionDefinitions.INPUT_SHORT_NAME, doc = "Input FASTQ file to be fixed. Default is STDIN.", optional=true)
     public File INPUT = null;
 
--- a/qmule/src/org/qcmg/qmule/FixCutAdapter.java
+++ b/qmule/src/org/qcmg/qmule/FixCutAdapter.java
@@ -16,17 +16,20 @@ import htsjdk.samtools.fastq.FastqWriterFactory;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
+import java.util.Date;
 
 public class FixCutAdapter extends CommandLineProgram {
-    static final String USAGE = "A Fastq file may contains empty seqence and base quality after cutAdapter,  Here A base N is set to empty sequence with 0 base quality value";
+    static final String USAGE = "A Fastq file may contains empty seqence after cutAdapter,  Here A base N is set to empty sequence with 0 base quality value";
     
     @Argument(shortName = StandardOptionDefinitions.INPUT_SHORT_NAME, doc = "Input Fastq file to be fixed.")
     public File INPUT;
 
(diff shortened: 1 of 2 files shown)


qmule/src/org/qcmg/qmule/FixCutAdapter.java
```diff
@@ -0,0 +1,101 @@
...
+import htsjdk.samtools.fastq.FastqWriterFactory;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.util.Date;
+
+public class FixCutAdapter extends CommandLineProgram {
+    static final String USAGE = "A Fastq file may contains empty seqence after cutAdapter,  Here A base N is set to empty sequence with 0 base quality value";
```
delocalizer, line 23: For clarity in the help; I suggest:
```
USAGE = "When cutadapt finds adapter sequence at the start of a read the default --action=trim results in empty sequence in the output. This tool replaces any empty sequence with a single base N with 0 base quality value.";
```

qmule/src/org/qcmg/qmule/FixCutAdapter.java
```diff
@@ -0,0 +1,101 @@
...
+    
+    @Argument(shortName = StandardOptionDefinitions.INPUT_SHORT_NAME, doc = "Input Fastq file to be fixed. Default is STDIN",optional=true)
```
delocalizer, line 25: In text visible to the user FASTQ should be either all upper or all lower `FASTQ` or `fastq`

qmule/src/org/qcmg/qmule/FixCutAdapter.java
```diff
@@ -0,0 +1,101 @@
...
+    public File INPUT = null;
+
+    @Argument(shortName = StandardOptionDefinitions.OUTPUT_SHORT_NAME, doc = "Where to write fixed fastq file.Default is STDOUT", optional=true)
```
delocalizer, line 28: suggest: 
```
"Where to write the fixed fastq records. Default is STDOUT"
```

delocalizer (commented):
Looks good, just some minor changes to the help text displayed to the user.

qmule/src/org/qcmg/qmule/FixCutAdapter.java
ChristinaXu2017: Our "cutadapt" version 1.9" doesn't support the "action" option, but provide option "--trim-n" or "--mask-adapter".  I will put this into the usage. Thanks.

qmule/src/org/qcmg/qmule/FixCutAdapter.java
ChristinaXu2017: fixed

delocalizer (changes requested):
`--trim-n` has nothing to do with the problem so referring to it in the help is just confusing. That option just trims `N` after adapter sequence is trimmed. Please use the text I suggested, or if you don't want to refer specifically to the newer tool options (which have been around for 4 years):
```
USAGE = "When cutadapt finds adapter sequence at the start of a read the default action of adapter sequence trimming results in empty sequence in the output. This tool replaces any empty sequence with a single base N with 0 base quality value.";
```

The actual problem is that the default tool action of adapter sequence trimming will leave empty sequence if the adapter is at the beginning of a read.

In new versions this can be toggled by changing the value of `--action=[trim|mask|...]`.

In very old versions like 1.9 that we use in our WDL there was no `--action` option, but the default action of the tool was to trim, and you could force it to mask by supplying `--mask-adapter`.

ChristinaXu2017:
yes, agree and fixed

delocalizer (approved):
looks good

PR #326, opened 2022-11-03
fix(q3tiledaligner): bug in comparator caused IllegalArgumentException

holmeso:
# Description

When sorting IntLongPair objects in a list, the comparator was ignoring the case where they could be equal. In certain circumstances this will throw an IllegalArgumentException "Comparison method violates its general contract!"

`TARecordUtil.getBestILPFromList` method had a lambda expression that sorted a list of `IntLongPair` objects.
It would not return 0 (the presumption being that you would never find 2 "equal" `IntLongPair` objects).

This caused the problem when comparing the following 2 objects:

```
IntLongPair ilp1 = new IntLongPair(1507329, 4611878433244640844l);
IntLongPair ilp2 = new IntLongPair(1507329, 4611878433244640646l);
```

`compare(ilp1, ilp2)` returns 1
`compare(ilp2, ilp1) `also returns 1
This violates the general contract.

The fix was to return 0 when 2 `IntLongPair` objects are deemed "equal" (as they are in this case)

Also changed here is the compareTo method in `BLATRecord` which now no longer throws a NPE when comparing records that have a `null` name. Some unit tests accompany this change.

## Type of change

# How Has This Been Tested?

Additional unit test has been added. Updated code runs (without error) against sequence that caused initial `IllegalArgumentException`

# Are WDL Updates Required?

No

# Checklist:

### DIFF
--- a/q3tiledaligner/test/au/edu/qimr/tiledaligner/util/TARecordUtilTest.java
+++ b/q3tiledaligner/test/au/edu/qimr/tiledaligner/util/TARecordUtilTest.java
@@ -2659,6 +2659,15 @@ public class TARecordUtilTest {
 		assertEquals(false, TARecordUtil.doGenomicPositionsOverlap(1994, 2121, 2011, 2129, 149));
 	}
 	
+	@Test
+	public void compareILPs() {
+		IntLongPair constant = new IntLongPair(2293761, 122046073075177l);
+		IntLongPair ilp1 = new IntLongPair(1507329, 4611878433244640844l);
+		IntLongPair ilp2 = new IntLongPair(1507329, 4611878433244640646l);
+		assertEquals(0, TARecordUtil.compareILPs(ilp1, ilp2, constant));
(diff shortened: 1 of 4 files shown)


q3tiledaligner/test/au/edu/qimr/tiledaligner/util/TARecordUtilTest.java
```diff
@@ -2659,6 +2659,15 @@ public void longsOverlapping2() {
 		assertEquals(false, TARecordUtil.doGenomicPositionsOverlap(1994, 2121, 2011, 2129, 149));
 	}
 	
+	@Test
```
delocalizer, line 2662: May I suggest a test each for the < and > case as well?

delocalizer (approved):
looks good

q3tiledaligner/test/au/edu/qimr/tiledaligner/util/TARecordUtilTest.java
holmeso: thanks for the suggestion - done

PR #336, opened 2023-07-14
qprofiler1 / qvisualise - handle new fastq header

holmeso:
fix(qvisualise/qprofiler): deal with new fastq header and make qvis more robust to deal with these

2 issues here, `qprofiler` was not aware of this new `fastq` header format and so each header became a unique instrument (all 28 million of them). `Qvisualise` would then run out of memory trying to deal
with them  all.   Have added code to properly deal with the new format,  in `qprofiler`, and
perhaps more importantly, added some limits to the size of collections that can be handled by
`qvisualise`.

New fastq header:
`@SRR14585604.8 A00805:41:HMJJWDRXX:1:1101:16929:1000 length=101`
which looks like the NCBI header as described [here](https://en.wikipedia.org/wiki/FASTQ_format#NCBI_Sequence_Read_Archive)

What `qprofiler` will now do for this is to count the number of spaces in the header. If there are 2, it will just send the middle part to be analysed. For any other number of spaces, the header will be treated as it was previously.

`qvisualise` will now run some checks on the input xml file size along with the supplied Xmx parameter, and log if it thinks a memory issue may be encountered.
It will also ignore any xml element that has more than 100000 items in it.

## Type of change

# How Has This Been Tested?

Additional unit test have been added.
Code has been run against offending fastq file and has produced satisfactory results

# Are WDL Updates Required?

Nope

# Checklist:


qprofiler/src/org/qcmg/qprofiler/fastq/FastqSummaryReport.java
```diff
@@ -292,18 +292,18 @@ void parseFiveElementHeaderWithSpaces(String [] params) {
 		// split by space
 		String [] firstElementParams = params[0].split(" ");
 		if (firstElementParams.length != 2) {
-			throw new UnsupportedOperationException("Incorrect header format encountered in parseFiveElementHeader. Expected '@ERR091788.3104 HSQ955_155:2:1101:13051:2071/2' but recieved: " + Arrays.deepToString(params));
+			throw new UnsupportedOperationException("Incorrect header format encountered in parseFiveElementHeader. Expected '@ERR091788.3104 HSQ955_155:2:1101:13051:2071/2' but received: " + Arrays.deepToString(params));
```
delocalizer, line 295: Here and ll 299, 306 the Exception messages are a bit confusing because they refer not to a generic format but to a specific example as 'Expected'; is there a better wording for these?

qvisualise/src/org/qcmg/qvisualise/QVisualise.java
```diff
@@ -27,54 +18,74 @@
...
-	
+
+	public static void examineFileSizeAndXmx(String inputFile) throws IOException {
+		long fileSize = Files.size(Paths.get(inputFile));
+		long xmxSize = Runtime.getRuntime().maxMemory();
+		logger.info("input file size: " + fileSize);
+		logger.info("supplied Xmx memory: " + xmxSize);
+		double ratio = (double)xmxSize / fileSize;
+		logger.info("memory / file size ratio: " + ratio);
+		if (ratio < 8) {
```
delocalizer, line 48: add some comment about this heuristic?

qvisualise/src/org/qcmg/qvisualise/util/QProfilerCollectionsUtils.java
```diff
@@ -138,41 +142,45 @@ public static Map<Integer, String> generatePercentagesMapFromElement(Element ele
...
-					if (tallyItemElement
-							.getAttribute("count").length() > 0) {
-						
-						final long count = Long.parseLong(tallyItemElement
-								.getAttribute("count"));
-						
-						map.put(isInteger ? (T) Integer.valueOf(tallyItemElement.getAttribute("value")) 
-								: (T) tallyItemElement.getAttribute("value") , new AtomicLong(count));
-												
+			if (tallyItemsNL.getLength() < 100000) {
```
delocalizer, line 152: magic number -> constant

qvisualise/src/org/qcmg/qvisualise/util/QProfilerCollectionsUtils.java
```diff
@@ -138,41 +142,45 @@ public static Map<Integer, String> generatePercentagesMapFromElement(Element ele
...
+									.getAttribute("count"));
+
+							map.put(isInteger ? (T) Integer.valueOf(tallyItemElement.getAttribute("value"))
+									: (T) tallyItemElement.getAttribute("value"), new AtomicLong(count));
+
+						}
 					}
 				}
+			} else {
+				log.warn(cycleElement.getTagName() + " has too many elements: " + tallyItemsNL.getLength() + " - will leave out of report");
```
delocalizer, line 169: and refer to the exceeded constant value here in the log message

delocalizer (approved):
Looks good. A couple of minor changes suggested.

delocalizer (approved):

PR #341, opened 2024-01-03
build(all projects): gradle updates, now works with vscode

holmeso:
# Description

Small changes to gradle build files that allows users to use vscode.
Use `implementation` rather that `api` in most projects when declaring dependencies.

# How Has This Been Tested?

Existing tests pass, existing build processes work

# Are WDL Updates Required?
no

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

PR #342, opened 2024-01-31
fix(qannotate): snpeff records were not being dealt with correctly

holmeso:
# Description

If a vcf record being annotated has more than 1 entry in the alt field, then nanno will split that record into 2 records.
eg.
`chr1 1000 A C,T .....`
will become
`chr1 1000 A C .....`
and
`chr1 1000 A T .....`

However, nanno was using the whole of the alt field in snpEff annotation files, which meant that there would not be a match from snpEff (just for records that had more than 1 entry in the alt field).

The fix involves splitting the 'alt' field when it contains more than one value, allowing the VCF record to be annotated to match against each of these values.

Other changes:
 - reverted a change made be the previous PR #341 that removed a build.last step
 - renamed ChePositionRefAlt.getName() to getRef()
 - added GATK GT field to nano output
 - added original_alt field to nanno output (useful when splitting 1/2 variants)
 - keep original GATK_AD values rather than manipulating them (original_alt field allows for this)
 - update Executor so that spaces in the classpath (hello IDEA) don't cause the Process to fall over
 

## Type of change

# How Has This Been Tested?

New unit tests, along with testing on cluster

# Are WDL Updates Required?

No

# Checklist:

### DIFF
--- a/q3vcftools/src/au/edu/qimr/vcftools/AmalgamatorGS.java
+++ b/q3vcftools/src/au/edu/qimr/vcftools/AmalgamatorGS.java
@@ -153,10 +153,10 @@ public class AmalgamatorGS {
 						gsAndAllOnlyCount++;
 					}
 				}
-				String scoreS = score + "/" + (4 + vcfFiles.length);
+				String scoreS = score + SLASH_STRING + (4 + vcfFiles.length);
 				scoreDist.computeIfAbsent(scoreS, v -> new AtomicInteger()).incrementAndGet();
 				ps.println(cp.getChromosome() + TAB + cp.getStartPosition() + TAB +cp.getRef() + TAB
-						+ cp.getAlt() + TAB + Arrays.stream(p.getLeft()).map(s -> null == s ? "./." : s).collect(Collectors.joining(TAB_STRING))
+						+ cp.getAlt() + TAB + Arrays.stream(p.getLeft()).map(s -> null == s ? MISSING_GT : s).collect(Collectors.joining(TAB_STRING))
 						+ TAB +  Arrays.stream(p.getRight()).map(s -> null == s ? MISSING_DATA_STRING : s).collect(Collectors.joining(TAB_STRING))
 						+ TAB +  Arrays.stream(missingACs).map(s -> null == s ? MISSING_DATA_STRING : s).collect(Collectors.joining(TAB_STRING))
 						+ TAB +  Arrays.stream(missingFTs).map(s -> null == s ? MISSING_DATA_STRING : s).collect(Collectors.joining(TAB_STRING))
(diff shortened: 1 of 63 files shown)


q3vcftools/src/au/edu/qimr/vcftools/AmalgamatorGS.java
```diff
@@ -169,7 +169,7 @@ private void writeOutput() throws FileNotFoundException {
 				}
 				String scoreS = score + "/" + (4 + vcfFiles.length);
 				scoreDist.computeIfAbsent(scoreS, v -> new AtomicInteger()).incrementAndGet();
-				ps.println(cp.getChromosome() + Constants.TAB + cp.getStartPosition() + Constants.TAB +cp.getName() + Constants.TAB 
+				ps.println(cp.getChromosome() + Constants.TAB + cp.getStartPosition() + Constants.TAB +cp.getRef() + Constants.TAB
 						+ cp.getAlt() + Constants.TAB + Arrays.stream(p.getLeft()).map(s -> null == s ? "./." : s).collect(Collectors.joining(Constants.TAB_STRING))
 						+ Constants.TAB +  Arrays.stream(p.getRight()).map(s -> null == s ? "." : s).collect(Collectors.joining(Constants.TAB_STRING))
 						+ Constants.TAB +  Arrays.stream(missingACs).map(s -> null == s ? "." : s).collect(Collectors.joining(Constants.TAB_STRING))
```
delocalizer, line 175: Minor query on consistency — is there a rationale for using `TAB` and `TAB_STRING` from `Constants` but not `SLASH_STRING`, `MISSING_DATA`, `MISSING_GT` etc. for the other string literals here?

q3vcftools/src/au/edu/qimr/vcftools/Classify.java
```diff
@@ -162,7 +162,7 @@ private void writeOutput(List<ChrPositionRefAlt> recs, String output) throws Fil
 			ps.println(header);
 			for (ChrPositionRefAlt cp : recs) {
 //				logger.info("writing out: " + cp.getName());
-				ps.println(cp.getChromosome() + Constants.TAB + cp.getStartPosition() + "\t.\t" + cp.getName() + Constants.TAB + cp.getAlt() + "\t.\t.\t.");
+				ps.println(cp.getChromosome() + Constants.TAB + cp.getStartPosition() + "\t.\t" + cp.getRef() + Constants.TAB + cp.getAlt() + "\t.\t.\t.");
```
delocalizer, line 165: While we're here, is this an opportunity to generate output in a nicer (more explicit) way, given that these look basically like VCF records? Essentially the same concatenation is used a couple more times below too.

delocalizer (approved):

q3vcftools/src/au/edu/qimr/vcftools/Classify.java
holmeso: added a method to ChrPositionUtils that takes a ChrPosition (and some other bits of info) and returns a Vcf string

q3vcftools/src/au/edu/qimr/vcftools/AmalgamatorGS.java
holmeso: Complete lack of consistency - using more Constatns now (even using static imports)

PR #347, opened 2024-02-22
feat(qannotate confidence mode): allow the user to specify the homopolymer cutoff

holmeso:
# Description

`qannotate's` `ConfidenceMode` was using a hard-coded value to determine if the `HOM` annotation should be applied to the filter field.
This PR allows the user to specify the homopolymer cutoff (via the `homCutoff` option), with the default value being the same as the previous hard-coded value.
This should mean that if you run the code as you would normally, then you will get identical results as before.
If you run with the `homCutoff` value set to something other than the default, you will get different results.

Some formatting has also been performed, and removal of user specific details from `qannotate's` `build.gradle` file.

## Type of change

# How Has This Been Tested?

Unit tests pass. Code has been run against vcf file, and results verified.

# Are WDL Updates Required?

No, assuming that the current defaults are acceptable, yes otherwise

# Checklist:

### DIFF
--- a/qannotate/src/au/edu/qimr/qannotate/modes/HomoplymersMode.java
+++ b/qannotate/src/au/edu/qimr/qannotate/modes/HomoplymersMode.java
@@ -1,8 +1,8 @@
 /**
  * © Copyright QIMR Berghofer Medical Research Institute 2014-2016.
- *
+ * <p>
  * This code is released under the terms outlined in the included LICENSE file.
-*/
+ */
 package au.edu.qimr.qannotate.modes;
 
 import java.io.File;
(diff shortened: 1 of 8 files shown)


qannotate/src/au/edu/qimr/qannotate/modes/HomoplymersMode.java
delocalizer, line 1: Can we fix the spelling of this class while we're here `HomopolymersMode`

delocalizer (approved):
Looks great — I wish I'd had this dial a while back! When I was benchmarking our pipeline recall and precision against a simulated truth set of variants I saw that most of our (very infrequent) false negative calls were due to HOM and MIN filters and did wonder if tuning could improve that.
https://genomeinfo.qimrberghofer.edu.au/svn/genomeinfo/adhoc/conradL/work_record/20190719_seq_depth/90%25.90x.README

qannotate/src/au/edu/qimr/qannotate/modes/HomoplymersMode.java
holmeso: yes we can - HomoplymersMode be gone!

PR #348, opened 2024-02-27
refactor(qannotate nanno): use long to store chr and position 

holmeso:
# Description

Chromosome  and position can be stored as a long (rather than a `ChrPosition` object) when interrogating annotation files.
Motivation for this change is that some of the annotation sources are large (dbNSFP has ~ 84M records, gnomad r3.1.2 ~750M records), and if the number of objects being created can be reduced, then garbage collectors have less to do, and applications should run faster.

To facilitate this change, an extra parameter has been added to the AnnotationSources json file called `chrStartsWithChr`. This field accepts a boolean value and indicates if the contigs within the annotation source file starts with the string "chr" eg. "chr1", "chr2", etc...)

# How Has This Been Tested?

Existing tests pass, has been run and compared against nanno output from regression tests.

# Are WDL Updates Required?

yes, the `imports/annotate/nannoCreateJson.wdl` wdl file will need to be updated and for each `NannoInput`, the `chrStartsWithChr` attribute will need to be added.

# Checklist:

### DIFF
--- a/qcommon/src/org/qcmg/common/util/ChrPositionUtils.java
+++ b/qcommon/src/org/qcmg/common/util/ChrPositionUtils.java
@@ -79,6 +79,7 @@ public class ChrPositionUtils {
             case "X" -> 23;
             case "Y" -> 24;
             case "M" -> 25;
+            case "MT" -> 25;
             default -> contigName.hashCode();
         };
     }
--- a/qcommon/src/org/qcmg/common/util/ChrPositionUtils.java
+++ b/qcommon/src/org/qcmg/common/util/ChrPositionUtils.java
@@ -31,6 +31,59 @@ public class ChrPositionUtils {
     }
 
     /**
+     * convert ChrPosition to a long. Will only examine the contig and start position
+     *
+     */
+    public static long convertContigAndPositionToLong(String contig, int position) {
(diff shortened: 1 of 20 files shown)


qcommon/src/org/qcmg/common/util/ChrPositionUtils.java
```diff
@@ -30,6 +30,59 @@ public static boolean doChrPositionsOverlap(ChrPosition a, ChrPosition b) {
...
+        }
+
+        if (contigName.length() > 3 && contigName.startsWith("chr")) {
+            return convertContigNameToInt(contigName.substring(3));
+        }
+
+        return switch (contigName) {
+            case "X" -> 23;
+            case "Y" -> 24;
+            case "M" -> 25;
```
delocalizer, line 81: handle MT (from chrMT) as well?

delocalizer (approved):
Looks good! I'd love to know what sort of performance improvement this gives.

qcommon/src/org/qcmg/common/util/ChrPositionUtils.java
holmeso: yep. It was left out because I think our version of the GRCh38 ref uses M rather than MT, but can do no harm having it in there and will no doubt save some errors in the future

PR #355, opened 2024-04-23
Update BamSummaryReport.java

holmeso:
fixed a bug introduced by my refactoring...


delocalizer (approved):
Sorry I missed that change of scope in review

holmeso:
> Sorry I missed that change of scope in review

Not at all, I shall modify my commit behaviour so as not to do whole file reformatting which makes spotting things like this near on impossible

PR #372, opened 2024-08-02
perf(qprofiler2): reduce calls to size on map

holmeso:
# Description

When profiling qprofiler2 running against a CRAM, I noticed that the producer thread was spending around 30% of its time calling the size method on the queue that it was populating (and that the consumer threads were ingesting from).
I have reduced the frequency of these calls to size, from every record to every 1 million records.

Also use a ConcurrentHashMap rather than a ConcurrentSkipListMap for the Tag tallies - much faster, and we can order at the point of displaying the data

And finally, some changes suggested by my IDE to make the code more up-to-date

# How Has This Been Tested?

Existing tests pass, and the updated ode has been profiled.

# Are WDL Updates Required?

nope

# Checklist:


newellf (approved):
Looks good, thanks.

